PR #6, opened 2023-09-28
Use enum for DriveOrder DriveType OrderDirection

phil-davis:
We already use `enum` for https://github.com/owncloud/ocis-php-sdk/blob/main/src/ResourceMetadata.php

Refactor to use `enum` for DriveOrder DriveType OrderDirection

This prevents the possibility of anyone passing an invalid value. If they do, PHP itself will complain straight away, we do not have to write code to check.

I looked at `README.md` and I don't think that the examples there need any changes. Methods like `listMyDrives` still take parameters specified like `DriveOrder::NAME` `OrderDirection::ASC` `DriveType::PERSONAL` - they are now `enum`s but they are specified the same way as when they were public constants.

### DIFF
--- a/src/DriveOrder.php
+++ b/src/DriveOrder.php
@@ -2,17 +2,8 @@
 
 namespace Owncloud\OcisPhpSdk;
 
-abstract class DriveOrder
+enum DriveOrder: string
 {
-    public const LASTMODIFIED = "lastModifiedDateTime";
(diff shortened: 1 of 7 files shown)


individual-it (changes requested):
good, deleted some code!

individual-it (approved):

PR #7, opened 2023-09-28
use webfinger

individual-it:
In more complex setups the service Url might not be known at the beginning but the webfinger service needs to be asked.
As the SDK expects a token to be present it only does authenticated requests to webfinger.

In this PR I'm introducing a new parameter to `$connectionConfig` to switch on the webfinger discovery.
so the object creation would be:
`$ocis = new Ocis($webfinger_url, $access_token->token, ['webfinger' => true]);`

The SDK then contacts the webfinger service, finds the first link with `"rel": "http://webfinger.owncloud/rel/server-instance"` and uses that as the service URL to do Graph & WebDAV requests.

The corresponding PR for moodle: https://github.com/owncloud/moodle-repository_ocis/pull/5

For the unit tests I've refactored the `$connectionConfig`, so that it can also take a guzzle(Mock). The reason is, that we need guzzle already in the constructor and so it does not work for the tests to call `setGuzzle`.

@phil-davis @amrita-shrestha @S-Panta @nirajacharya2 is the design using `['webfinger' => true]` in `$connectionConfig` good, or should we introduce an other parameter to `Ocis::__constructor`?

@micbar 
1. would that approach for webfinger discovery work in bigger instances?
2. I set the `resource` parameter in the webfinger URL to `acct:me@<host of the iss in the token>`. Is that correct or does it even matter?

### DIFF
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -30,7 +30,7 @@ class Drive
     private string $webDavUrl = '';
 
     /**
-     * @phpstan-var array{'headers'?:array<string, mixed>, 'verify'?:bool}
+     * @phpstan-var array{'headers'?:array<string, mixed>, 'verify'?:bool, 'webfinger'?:bool, 'guzzle'?:\GuzzleHttp\Client}
      */
     private array $connectionConfig;
     private Configuration $graphApiConfig;
@@ -39,7 +39,7 @@ class Drive
     /**
      * @ignore The developer using the SDK does not need to create drives manually, but should use the Ocis class
      *         to get or create drives, so this constructor should not be listed in the documentation.
-     * @phpstan-param array{'headers'?:array<string, mixed>, 'verify'?:bool} $connectionConfig
+     * @phpstan-param array{'headers'?:array<string, mixed>, 'verify'?:bool, 'webfinger'?:bool, 'guzzle'?:\GuzzleHttp\Client} $connectionConfig
      * @throws \InvalidArgumentException
      */
     public function __construct(
--- a/tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
+++ b/tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace unit\Owncloud\OcisPhpSdk;
+
+use GuzzleHttp\Client;
+use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
+use Owncloud\OcisPhpSdk\Ocis;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\StreamInterface;
+
+class OcisWebfingerTest extends TestCase
+{
+    private function getValidToken(): string
+    {
+        $tokenHeader = [
+            "alg" => "PS256",
+            "kid" => "private-key",
+            "typ" => "JWT"
+        ];
+        $tokenPayload = [
+            "iss" => "https://sso.example.com",
+        ];
+        return base64_encode((string)json_encode($tokenHeader)) . "." .
+            base64_encode((string)json_encode($tokenPayload)) .
+            ".signatureDoesNotMatter";
+    }
+    private function getGuzzleMock(?string $responseContent = null): MockObject
+    {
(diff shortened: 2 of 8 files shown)


src/Drive.php
```diff
@@ -17,12 +22,12 @@ class Drive
     private string $webDavUrl = '';
 
     /**
-     * @phpstan-var array{'headers'?:array<string, mixed>, 'verify'?:bool}
+     * @phpstan-var array{'headers'?:array<string, mixed>, 'verify'?:bool, 'webfinger'?:bool, 'guzzle'?:Client}
```
phil-davis, line 25:
```suggestion
     * @phpstan-var array{'headers'?:array<string, mixed>, 'verify'?:bool, 'webfinger'?:bool, 'guzzle'?:GuzzleHttp\Client}
```

Maybe you can write the full thing here, and not have to add `use GuzzleHttp\Client;` above?

tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->assertSame("https://abc.drive.example.com", $ocis->getServiceUrl());
+    }
+
+    public function invalidTokenProvider(): array
+    {
+        return [
+            ["onlyHeaderNoPayload"],
+            ["header.butPäylöädNötBäs€64"],
+            ["header.cGF5bG9hZElzTm90SlNPTgo="], // payload is not JSON
+            ["header.InBheWxvYWRJc05vdEFuT2JqZWN0Igo="], // payload is not an JSON object
```
phil-davis, line 66:
```suggestion
            ["header.InBheWxvYWRJc05vdEFuT2JqZWN0Igo="], // payload is not a JSON object
```

tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
```diff
@@ -0,0 +1,115 @@
...
+            ["header.eyJrZXkiOiAidmFsdWUifQo="], // payload is a JSON object, but does not contain 'iss' key
+            ["header.eyJpc3MiOiBudWxsfQo="], // payload contains 'iss' key but the values is null
```
phil-davis, line 68:
```suggestion
            ["header.eyJpc3MiOiBudWxsfQo="], // payload contains 'iss' key but the value is null
```

tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
```diff
@@ -0,0 +1,115 @@
...
+            ["header.eyJpc3MiOiAiaHR0cHM6Ly8ifQo="], // payload contains 'iss' key but the values is 'https://'
```
phil-davis, line 69:
```suggestion
            ["header.eyJpc3MiOiAiaHR0cHM6Ly8ifQo="], // payload contains 'iss' key but the value is 'https://'
```

tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
```diff
@@ -0,0 +1,115 @@
...
+            ["header.eyJpc3MiOiAiaG9zdCJ9Cg=="], // payload contains 'iss' key but the values is 'host'
```
phil-davis, line 70:
```suggestion
            ["header.eyJpc3MiOiAiaG9zdCJ9Cg=="], // payload contains 'iss' key but the value is 'host'
```

tests/unit/Owncloud/OcisSdkPhp/OcisWebfingerTest.php
```diff
@@ -0,0 +1,115 @@
...
+            ["header.eyJpc3MiOiAiaHR0cHM6Ly86OTAwMCJ9Cg=="], // payload contains 'iss' key but the values is 'https://:9000'
```
phil-davis, line 71:
```suggestion
            ["header.eyJpc3MiOiAiaHR0cHM6Ly86OTAwMCJ9Cg=="], // payload contains 'iss' key but the value is 'https://:9000'
```

phil-davis (commented):
Looking good so far.

amrita-shrestha (commented):
lgtm

amrita-shrestha (approved):

phil-davis (approved):

PR #12, opened 2023-10-05
improve exception handling & documentation

individual-it:
1. document in more details which exceptions are thrown
2. throw better exceptions in some cases

### DIFF
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -6,7 +6,11 @@ use DateTime;
 use OpenAPI\Client\Model\Drive as ApiDrive;
 use OpenAPI\Client\Model\DriveItem;
 use OpenAPI\Client\Model\Quota;
+use Owncloud\OcisPhpSdk\Exception\BadRequestException;
 use Owncloud\OcisPhpSdk\Exception\ExceptionHelper;
+use Owncloud\OcisPhpSdk\Exception\ForbiddenException;
+use Owncloud\OcisPhpSdk\Exception\NotFoundException;
+use Owncloud\OcisPhpSdk\Exception\UnauthorizedException;
 use Sabre\HTTP\ClientException as SabreClientException;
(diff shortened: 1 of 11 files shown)


src/Ocis.php
```diff
@@ -182,6 +191,11 @@ public function listAllDrives(
      * Get all drives that the current user is a regular member of
      *
      * @return array<Drive>
+     *
+     * @throws BadRequestException
+     * @throws ForbiddenException
+     * @throws NotFoundException
+     * @throws UnauthorizedException
      * @throws \Exception
```
phil-davis, line 199: Do we still need this general Exception line?

src/Drive.php
```diff
@@ -277,6 +294,10 @@ public function uploadFile(string $path, string $content): bool
      * @param resource|string|null $resource file resource pointing to the file to be uploaded
      *
      * @return bool
+     * @throws BadRequestException
+     * @throws ForbiddenException
+     * @throws NotFoundException
+     * @throws UnauthorizedException
      * @throws \Exception
```
phil-davis, line 301: Do we still need lines like this?
Are there other "unusual" exceptions that are known to be possible?

src/Ocis.php
```diff
@@ -248,8 +267,10 @@ public function getDriveById(string $driveId): Drive
      * @param int $quota in bytes
      * @return Drive
      * @throws \Exception
```
phil-davis, line 269: Maybe this line can be deleted?

phil-davis (approved):

src/Drive.php
individual-it: we throw the \Exception ourself in this case in line 308. I could change it to `InvalidParameterException`

individual-it:
@phil-davis I've refactored the code, so that it never throws general Exceptions

phil-davis (approved):

PR #13, opened 2023-10-06
improve and publish documentation

individual-it:
1. improve documentation of classes and functions
2. improve exception messages
3. generate docs using phpdoc.org/
4. publish docs to https://owncloud.dev/ocis-php-sdk/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ This SDK allows you to interact with [ownCloud Infinite Scale (oCIS)](https://gi
 ## Documentation
 You can find a rendered version of the [API documentation](https://owncloud.dev/ocis-php-sdk/) in our dev docs.
 
+To render the documentation locally, use the [phpDocumentor](https://www.phpdoc.org/) to run it in the local repo. E.g.:
+```
+docker run --rm -v ${PWD}:/data phpdoc/phpdoc:3
+```
+
+After that you will find the documentation inside the `docs` folder.
+
 :exclamation: This SDK is still under heavy development and is not yet ready for production use, the API might change!
 
 ## Getting started
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # ocis-php-sdk
 This SDK allows you to interact with [ownCloud Infinite Scale (oCIS)](https://github.com/owncloud/ocis/) storage using PHP.
 
+## Documentation
+You can find a rendered version of the [API documentation](https://owncloud.dev/ocis-php-sdk/) in our dev docs.
+
 :exclamation: This SDK is still under heavy development and is not yet ready for production use, the API might change!
 
 ## Getting started
@@ -52,7 +55,7 @@ $drives[0]->uploadFile("/documents/myfile.txt", "Hello World!");
 $resources = $drives[0]->listResources("/documents");
 ```
 
-### Notifications
+## Notifications
 Notifications can be listed using the `listNotifications` method, which will return an array of `Notification` objects representing all active notifications.
 
 The `Notification` object can retrieve details of the corresponding notification and mark it as read (delete).
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -20,6 +20,9 @@ use Owncloud\OcisPhpSdk\Exception\UnauthorizedException;
 use Sabre\HTTP\ClientException as SabreClientException;
 use Sabre\HTTP\ClientHttpException as SabreClientHttpException;
 
+/**
+ * Class representing a single drive/space in ownCloud Infinite Scale
+ */
 class Drive
 {
     private ApiDrive $apiDrive;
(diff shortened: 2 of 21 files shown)


README.md
```diff
@@ -1,6 +1,9 @@
 # ocis-php-sdk
 This SDK allows you to interact with [ownCloud Infinite Scale (oCIS)](https://github.com/owncloud/ocis/) storage using PHP.
 
+## Documentation
+You can find a rendered version of the [API documentation](https://owncloud.dev/ocis-php-sdk/) in our dev docs.
+
```
amrita-shrestha, line 6: Is there any documentation to locally serve documentation

src/Drive.php
```diff
@@ -227,17 +253,18 @@ public function getFile(string $path)
     }
 
     /**
-     * get file as a file resource
+     * Download the file at the given path in the current drive and return the stream
```
amrita-shrestha, line 256:
```suggestion
     * Download the file as a stream at the given path in the current drive
```

src/Drive.php
```diff
@@ -258,18 +285,25 @@ public function createFolder(string $path): bool
...
+     */
     public function getResourceMetadataById(string $id): \stdClass
     {
-        throw new \Exception("This function is not implemented yet.");
+        throw new \Exception("This function is not implemented yet! Place, name and signature of the function might change!");
     }
 
     /**
-     * update file content if the file already exists
+     * Update the file content
```
amrita-shrestha, line 305: this method creates new file  if file doesn't exist and updates content if the file exists

README.md
individual-it: good point, added some description in 696f213

src/Drive.php
individual-it: I think the function will return the stream and the file will only be downloaded when the stream is read

phil-davis:
@individual-it needs a rebase to resolve conflicts

individual-it:
@phil-davis done

src/Drive.php
```diff
@@ -287,17 +307,18 @@ public function getFile(string $path)
     }
 
     /**
-     * get file as a file resource
+     * Download the file at the given path in the current drive and return the stream
```
phil-davis, line 310:
```suggestion
     * Return a stream resource for the file at the given path in the current drive. The content of the file can be read from the stream.
```

Maybe text like this?

src/Drive.php
```diff
@@ -380,7 +409,7 @@ public function uploadFileStream(string $path, $resource): bool
     }
 
     /**
-     * delete resource
+     * Delete (move to trash-bin) the Ocis resource (file or folder) at the given path in the current drive
```
phil-davis, line 412: Does every ocis have a trash bin?
Or is it possible to configure ocis without a trash bin?

src/DriveOrder.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Owncloud\OcisPhpSdk;
 
+/**
+ * Possible order values for sorting drives
```
phil-davis, line 6:
```suggestion
 * Possible order values for sorting a list of drives
```

src/Exception/ExceptionHelper.php
```diff
@@ -7,8 +7,16 @@
...
 use Sabre\HTTP\ClientHttpException as SabreClientHttpException;
 
+/**
+ * @ignore This is only used for internal purposes and should not show up in the documentation
+ * Helper class to help with handling different exceptions thrown by dependencies
+ */
 class ExceptionHelper
 {
+    /**
+     * Takes an exception thrown by a dependency and returns an custom exception
```
phil-davis, line 17:
```suggestion
     * Takes an exception thrown by a dependency and returns a custom exception
```

src/Exception/InvalidResponseException.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Owncloud\OcisPhpSdk\Exception;
 
+/**
+ * Exception thrown in the case the ownCloud Infinite Scale server returns an invalid response.
```
phil-davis, line 6:
```suggestion
 * Exception thrown if the ownCloud Infinite Scale server returns an invalid response.
```

src/Ocis.php
```diff
@@ -57,20 +60,24 @@ public function __construct(
...
+     * @ignore This function is mainly for unit tests and should not be shown in the documentation
+     */
     public function setGuzzle(Client $guzzle): void
     {
         $this->guzzle = $guzzle;
     }
 
     /**
      * @param array<mixed> $connectionConfig
+     * @ignore This function is used for internal purposes only and should not be shown in the documentation. The function is public to make it testable and because its used also from other classes.
```
phil-davis, line 80:
```suggestion
     * @ignore This function is used for internal purposes only and should not be shown in the documentation. The function is public to make it testable and because its also used from other classes.
```

src/OrderDirection.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Owncloud\OcisPhpSdk;
 
+/**
+ * Possible order directions values for sorting responses
```
phil-davis, line 6:
```suggestion
 * Possible order direction values for sorting responses
```

src/Drive.php
individual-it: I think every storage driver is supposed to implement a trash
`After deleting a node the storage allows listing the deleted nodes and has an undo mechanism for them.` from:
https://owncloud.dev/ocis/storage/storagedrivers/#trash-persistence

phil-davis (approved):

PR #15, opened 2023-10-10
test coverage

individual-it:
Enable sonar cloud analysis
I believe the coverage information we will be able to see after merging to main

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,32 @@
+# Organization and project keys are displayed in the right sidebar of the project homepage
+sonar.organization=owncloud-1
+sonar.projectKey=owncloud_ocis-php-sdk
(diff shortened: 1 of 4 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,32 @@
+# Organization and project keys are displayed in the right sidebar of the project homepage
+sonar.organization=owncloud-1
+sonar.projectKey=owncloud_ocis-php-sdk
+sonar.projectVersion=1.6.0
```
phil-davis, line 4:
```suggestion
sonar.projectVersion=0.1.0
```

I think that we want to start with some low version, until we actually start making releases.

phil-davis (approved):
Approving - only the version number needs adjusting before merge.

PR #20, opened 2023-10-11
improve integration tests docker

individual-it:
1. use `ocis.owncloud.test` and `keycloak.owncloud.test` instead of ocis.sdk.test
2. don't use `host-gateway` to avoid firewall issues
3. add healthchecks for every service
4. start depending service only when the healthcheck of the previous passed, to make sure the tests are not executed too early
5. show more log messages for ocis
6. added instructions how to run integration tests into the README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ test-php-unit: vendor/bin/phpunit
 .PHONY: test-php-integration
 test-php-integration:             ## Run php integration tests
 test-php-integration: run-ocis-with-keycloak
+	composer install
 	$(PHPUNIT) --configuration ./phpunit.xml --testsuite integration
 	$(MAKE) docker-clean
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -62,6 +63,7 @@ docker-clean:
 clean:
 	rm -f composer.lock .php-cs-fixer.cache .phpunit.result.cache
 	rm -Rf vendor
+	cd tests/integration && $(DCO) down -v --remove-orphans --rmi local
```
amrita-shrestha, line 66: any reason to add `--rmi local`

README.md
```diff
@@ -183,3 +183,20 @@ If you're working in a development environment where you might need to bypass SS
...
 
 To test, simply open a browser and head to http://url-of-this-file.
+
+
+## Development
+
+### Integration tests
+The integration tests start a full oCIS server with keycloak and other services using docker.
+To run the tests locally
+1. Install and setup `docker` and `docker compose`.
```
amrita-shrestha, line 193: IMO better if we specified version 2

Makefile
individual-it: so that when we run `clean` any image that was build will be deleted in case something changed

README.md
individual-it: done

amrita-shrestha (approved):
lgtm

PR #40, opened 2023-10-20
reduce complexity of getNotifications

### LINKED ISSUE #31: reduce complexity of getNotifications
Sonarcloud complains that the function is too complex to easily understand: https://sonarcloud.io/project/issues?impactSeverities=HIGH&resolved=false&id=owncloud_ocis-php-sdk&open=AYsY7u6LwCB8P94PULzl
Maybe we can reduce the nested if-else layers or put the validity checks in a separate private function

individual-it:
I hope this will make sonarcloud think that the function is now easier to understand.
I had to define all the parameters into single variables and declare the details of the array to make phpstan happy

fixes #31

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -526,6 +526,23 @@ class Ocis
     }
 
     /**
+     * @param array<mixed> $ocsResponse
+     * @return bool
+     */
+    private function isNotificationResponseValid(array $ocsResponse): bool
+    {
+        return (
+            is_array($ocsResponse) &&
+            isset($ocsResponse['ocs']) &&
+            array_key_exists('data', $ocsResponse['ocs']) &&
+            (
+                is_array($ocsResponse['ocs']['data']) ||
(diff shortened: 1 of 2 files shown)


src/Ocis.php
```diff
@@ -548,72 +564,75 @@ public function getNotifications(): array
...
+         *              subjectRich?: ?string,
+         *              message?: ?string,
+         *              messageRich?: ?string,
+         *              messageRichParameters?:array{int, mixed}
+         *          }
+         *      }
+         *  }
+         *} $ocsResponse
+         */
+        $ocsResponse = (array)json_decode($content, true);
```
phil-davis, line 588: The (array) cast should work here, even if the result of `json_decode` was a scalar (string, int, boolean) or null. It will produce an  array with one entry. That will be fine to pass to `isNotificationResponseValid` which will examine the array and work out that it is not a valid response.

phil-davis (approved):

PR #45, opened 2023-10-30
check the connection config dynamically

### LINKED ISSUE #35: refactor isConnectionConfigValid, so that it checks the types by the values int the validConnectionConfigKeys array
there is an array with valid keys, we can use the values to check the type of the values passed in by `connectionConfig`, similar as done for the `guzzle` value.
That should reduce the complexity and the amount of return statements in that function

see also: https://sonarcloud.io/project/issues?open=AYsY7u6LwCB8P94PULzk&id=owncloud_ocis-php-sdk&tab=code

individual-it:
with this check we don't need to write an `if` statement for every key, but can just give the check in the value of `$validConnectionConfigKeys`
this should also fix the sonar issue with too many return statements, see https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&resolved=false&id=owncloud_ocis-php-sdk&open=AYsY7u6LwCB8P94PULzk

fixes #35

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -103,7 +103,7 @@ class Ocis
      * Helper function to check if the variable is a guzzle client
      * we need this because we want to call the check with call_user_func
      *
-     * @phpstan-ignore-next-line phpstan does not understand that this mdethod was called via call_user_func
+     * @phpstan-ignore-next-line phpstan does not understand that this method was called via call_user_func
      */
     private static function isGuzzleClient(mixed $guzzle): bool
     {
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -103,7 +103,7 @@ class Ocis
      * Helper function to check if the variable is a guzzle client
      * we need this because we want to call the check with call_user_func
      *
-     * @phpstan-ignore-next-line phpstan does not understand that the was called via call_user_func
+     * @phpstan-ignore-next-line phpstan does not understand that this mdethod was called via call_user_func
      */
     private static function isGuzzleClient(mixed $guzzle): bool
     {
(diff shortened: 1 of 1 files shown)


src/Ocis.php
```diff
@@ -99,6 +99,17 @@ public function getServiceUrl(): string
         return $this->serviceUrl;
     }
 
+    /**
+     * Helper function to check if the variable is a guzzle client
+     * we need this because we want to call the check with call_user_func
+     *
+     * @phpstan-ignore-next-line phpstan does not understand that the was called via call_user_func
```
phil-davis, line 106:
```suggestion
     * @phpstan-ignore-next-line phpstan does not understand that this mdethod was called via call_user_func
```

src/Ocis.php
```diff
@@ -107,30 +118,18 @@ public function getServiceUrl(): string
...
-        if (
-            array_key_exists('guzzle', $connectionConfig) &&
-            !($connectionConfig['guzzle'] instanceof $validConnectionConfigKeys['guzzle'])
-        ) {
-            return false;
-        }
-        if (array_key_exists('headers', $connectionConfig)) {
-            if (!is_array($connectionConfig['headers'])) {
+
+            // @phpstan-ignore-next-line phpstan does not understand that the `$check` is callable
```
phil-davis, line 131:
```suggestion
            // @phpstan-ignore-next-line phpstan does not understand that the `$validConnectionConfigKeys` array has values that are callable
```

src/Ocis.php
```diff
@@ -99,6 +99,17 @@ public function getServiceUrl(): string
         return $this->serviceUrl;
     }
 
+    /**
+     * Helper function to check if the variable is a guzzle client
+     * we need this because we want to call the check with call_user_func
+     *
+     * @phpstan-ignore-next-line phpstan does not understand that this mdethod was called via call_user_func
```
phil-davis, line 106:
```suggestion
     * @phpstan-ignore-next-line phpstan does not understand that this method was called via call_user_func
```

phil-davis (approved):

PR #54, opened 2023-10-30
reduce amount of parameters

individual-it:
this makes sonar cloud happy, see https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&resolved=false&id=owncloud_ocis-php-sdk&open=AYsY7u7CwCB8P94PUL0R&tab=code
but I'm not sure if we should do that, feels like having less control over the types


amrita-shrestha (approved):
lgtm

PR #56, opened 2023-10-31
adding create groups

nirajacharya2:
implementing create group

fixes : https://github.com/owncloud/ocis-php-sdk/issues/63

better to merge [this](https://github.com/owncloud/ocis-php-sdk/pull/66) first so the CI passes

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -24,6 +24,8 @@ use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
 use Sabre\HTTP\ResponseInterface;
 use stdClass;
 use OpenAPI\Client\Api\GroupsApi;
+use Owncloud\OcisPhpSdk\Group;
+use OpenAPI\Client\Model\Group as OpenAPIGrop;
 
 /**
  * Basic class to establish the connection to an ownCloud Infinite Scale instance
@@ -752,4 +754,35 @@ class Ocis
         }
         return $notifications;
     }
+
+    /**
+    * Create a new group (if the user has the permission to do so)
+    *
+    * @param string $groupName
+    * @param string $description
+    * @return void
+    * @throws BadRequestException
+    * @throws ForbiddenException
+    * @throws NotFoundException
+    * @throws UnauthorizedException
+    * @throws \InvalidArgumentException
+    * @throws InvalidResponseException
+    * @throws HttpException
+    */
+    public function createGroup(string $groupName, string $description = "")
+    {
+        $apiInstance = new GroupsApi($this->guzzle, $this->graphApiConfig);
+        $group = new OpenAPIGrop(["display_name" => $groupName, "description" => $description]);
+        try {
+            $newlyCreatedGroup = $apiInstance->createGroup($group);
+        } catch (ApiException $e) {
+            throw ExceptionHelper::getHttpErrorException($e);
+        }
+        if ($newlyCreatedGroup instanceof OdataError) {
+            throw new InvalidResponseException(
+                "createGroup returned an OdataError - " . $newlyCreatedGroup->getError()
+            );
+        }
+    }
+
 }
--- a/tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
@@ -155,8 +155,10 @@ class OcisTest extends OcisPhpSdkTestCase
     {
         $token = $this->getAccessToken("admin", "admin");
         $ocis = new Ocis($this->ocisUrl, $token, ["verify" => false]);
-        $ocis->createGroup("radio-lovers", "sss");
         $groups = $ocis->getGroups(search: $searchText, orderBy: $orderDirection);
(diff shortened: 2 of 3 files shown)


tests/integration/Owncloud/OcisSdkPhp/OcisTest.php
```diff
@@ -138,15 +138,43 @@ public function testCreateDriveInvalidQuota(int $quota): void
...
+     */
+    public function testCreateGroup($groupName, $groupDescription)
+    {
+        $token = $this->getAccessToken('admin', 'admin');
+        $ocis = new Ocis($this->ocisUrl, $token, ['verify' => false]);
+        $ocis->createGroup($groupName, $groupDescription);
+        $this->markTestSkipped('https://github.com/owncloud/ocis-php-sdk/issues/67');
+        // $this->assertInstanceOf(Group::class, $newGroup);
+        // $this->assertSame($newGroup->getDisplayName(),$groupName);
+        // $this->assertSame($newGroup->getDescription(),$groupDescription);
```
individual-it, line 167: that should not be a problem, because oDataError is not catch in the createGroup function

src/Ocis.php
```diff
@@ -22,6 +22,8 @@
 use Sabre\HTTP\ResponseInterface;
 use stdClass;
 use OpenAPI\Client\Api\GroupsApi;
+use Owncloud\OcisPhpSdk\Group;
+use OpenAPI\Client\Model\Group as libGroup;
```
individual-it, line 26:
```suggestion
use OpenAPI\Client\Model\Group as OpenAPIGroup;
```

individual-it (changes requested):

tests/integration/Owncloud/OcisSdkPhp/OcisTest.php
```diff
@@ -139,14 +139,19 @@ public function testCreateDriveInvalidQuota(int $quota): void
...
 
+    /**
+     * @return void
+     */
     public function testGetGroups(): void
     {
         $token = $this->getAccessToken('admin', 'admin');
         $ocis = new Ocis($this->ocisUrl, $token, ['verify' => false]);
+        $ocis->createGroup("philosophy-haters", "sss");
+        $ocis->createGroup("physics-lovers", "sss");
```
individual-it, line 150: We should move this to the `setUp` function, or have the `createGroup` calls in every test, otherwise we cannot run some tests independently of others and the order of tests cannot be changed.
In future we need to be able to delete groups, and then we can have a proper cleanuo

individual-it (changes requested):

amrita-shrestha (changes requested):
https status code issue has been resloved https://github.com/owncloud/ocis-php-sdk/issues/67#issuecomment-1813857984

tests/integration/Owncloud/OcisSdkPhp/OcisTest.php
S-Panta: If this is kept in `setup` , then ocis should be used as precondition for test since ocis ffirst need to get initialized. Also, we haven't implemented deletegroup, creating group in every step would mean that every assertion steps would have more group related information in it and it would fail when we implement delete group. We can use `createGroup` in every step once we implement the delete group.

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -130,8 +139,8 @@ public function testGetGroupSerch(string $searchText, int $count, array $groupDi
     public function orderDirection(): array
     {
         return [
-            [OrderDirection::ASC, "ph", ["philosophy-haters", "physics-lovers"]],
-            [OrderDirection::DESC, "ph", ["physics-lovers", "philosophy-haters"]],
+            [OrderDirection::ASC, "ph", ["philosophy-haters", "physics-lovers"]]
+//            [OrderDirection::DESC, "ph", ["physics-lovers", "philosophy-haters"]], // first implement cleanup group
```
individual-it, line 143: why does this need the cleanup?

S-Panta: Sending `POST` request twice without cleanup results in conflict group creation

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -146,10 +155,8 @@ public function testGetGroupSort(OrderDirection $orderDirection, string $searchT
     {
         $token = $this->getAccessToken("admin", "admin");
         $ocis = new Ocis($this->ocisUrl, $token, ["verify" => false]);
-        $groups = $ocis->getGroups(search: "user", orderBy: $orderDirection);
-        if(count($groups) <= 0) {
-            $this->markTestSkipped("no groups created");
-        }
+        $ocis->createGroup("radio-lovers", "sss");
```
individual-it, line 158: why creating this group?

individual-it (approved):

amrita-shrestha (approved):
lgtm

PR #58, opened 2023-11-03
implementing adding user to group

nirajacharya2:
implementing adding user to group

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -2,8 +2,19 @@
 
 namespace Owncloud\OcisPhpSdk;
 
+use GuzzleHttp\Client;
+use OpenAPI\Client\Api\GroupApi;
+use OpenAPI\Client\Configuration;
 use OpenAPI\Client\Model\Group as OpenApiGroup;
+use OpenAPI\Client\Model\MemberReference;
 use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
+use Owncloud\OcisPhpSdk\Exception\BadRequestException;
+use Owncloud\OcisPhpSdk\Exception\ExceptionHelper;
+use Owncloud\OcisPhpSdk\Exception\ForbiddenException;
+use Owncloud\OcisPhpSdk\Exception\HttpException;
+use Owncloud\OcisPhpSdk\Exception\NotFoundException;
+use Owncloud\OcisPhpSdk\Exception\UnauthorizedException;
+use OpenAPI\Client\ApiException;
 
 class Group
 {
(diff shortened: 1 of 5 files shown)


phil-davis:
@nirajacharya2  please rebase this so it "sees" the current master license.

src/Group.php
```diff
@@ -83,5 +121,18 @@ public function getMembers(): array
     {
         return $this->members;
     }
-
+    /**
```
amrita-shrestha, line 124:
```suggestion

    /**
```

src/Group.php
```diff
@@ -18,11 +22,38 @@ class Group
...
+     * @param string $accessToken
      */
-    public function __construct(OpenApiGroup $openApiGroup)
-    {
+    public function __construct(
+        OpenApiGroup $openApiGroup,
+        string $serviceUrl = "",
+        array $connectionConfig = [],
+        string &$accessToken = ""
+    ) {
```
individual-it, line 56: I think these parameters are not optional (except of `connectionConfig`)
```suggestion
    public function __construct(
        OpenApiGroup $openApiGroup,
        string $serviceUrl,
        array $connectionConfig,
        string &$accessToken
    ) {
```

src/Group.php
```diff
@@ -83,5 +121,18 @@ public function getMembers(): array
     {
         return $this->members;
     }
-
+    /**
+     * @param User $user
+     * @return void
+     */
+    public function addUserToGroup($user)
```
individual-it, line 128: If possible add the types into the function declaration
```suggestion
    public function addUserToGroup(User $user): void
```

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -89,6 +89,30 @@ public function testGetGroups(): void
             $this->assertIsArray($group->getMembers());
         }
     }
+
+    /**
+     * @return void
+     */
+    public function testAddUserAndGroupsExpand(): void
```
individual-it, line 96: what do you want to test here? adding user or getting group and expanding the members?
If you want to test both, write two separate tests

src/Group.php
individual-it, line 128: we know its about adding a user to a group, because we are using the Group class
```suggestion
    public function addUser($user)
```

src/Group.php
```diff
@@ -83,5 +121,18 @@ public function getMembers(): array
...
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        $memberRef = new MemberReference(
+            [
+                "at_odata_id" => $this->graphApiConfig->getHost(). "/v1.0/users/" . $user->getId()
+            ]
+        );
+        $apiInstance->addMember($this->getId(), $memberRef);
```
individual-it, line 136: catch the ApiException as we do in other places

individual-it (changes requested):
w

individual-it (approved):

PR #61, opened 2023-11-06
implementing sharing functionality

individual-it:
Implementing sharing of resources.

Sharing (invite) can happen on any resource.
Before the invite we need recipients (groups or users) and a role. The possible roles are retrieved from the item `"@libre.graph.permissions.roles.allowedValues"` in [/v1beta1/drives/{drive-id}/items/{item-id}/permissions](https://owncloud.dev/libre-graph-api/#/drives.permissions/ListPermissions)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ for ($roles as $role) {
 }
 
 // find all users with a specific surname
-$users = $ocis->listUsers("gurung");
+$users = $ocis->getUsers("gurung");
 
 // share the resource with the users
 $resources[0]->invite($users, $editorRole);
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Notifications can be listed using the `listNotifications` method, which will ret
 The `Notification` object can retrieve details of the corresponding notification and mark it as read (delete).
 
 ## Sharing
-Given the correct permissions, an `OcisResouce` can be shared with a group or a user. To define the access permissions of the receiver the every share has to set `SharingRole`(s).
+Given the correct permissions, an `OcisResource` can be shared with a group or a user. To define the access permissions of the receiver every share has to set `SharingRole`(s).
 
 ```php
 // get the resources of a subfolder inside a drive
--- a/README.md
+++ b/README.md
@@ -74,6 +74,31 @@ Notifications can be listed using the `listNotifications` method, which will ret
 
 The `Notification` object can retrieve details of the corresponding notification and mark it as read (delete).
 
+## Sharing
+Given the correct permissions, an `OcisResouce` can be shared with a group or a user. To define the access permissions of the receiver the every share has to set `SharingRole`(s).
+
+```php
+// get the resources of a subfolder inside a drive
+$resources = $drive->listResources("/documents");
+
+// get all roles that are possible for that particular resource
+$roles = $resources[0]->getRoles();
+
+// find the role that is allowed to read and write the shared file or folder 
+for ($roles as $role) {
+    if ($role->getDisplayName() === 'Editor') {
+        $editorRole = $role;
+        break;
+    }
+}
+
+// find all users with a specific surname
+$users = $ocis->listUsers("gurung");
+
+// share the resource with the users
+$resources[0]->invite($users, $editorRole);
+```
+
 ## Requirements
 - PHP 8.1 or higher
 - oCIS 4.0.0 or higher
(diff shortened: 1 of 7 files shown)


individual-it:
CC @phil-davis @amrita-shrestha @S-Panta @nirajacharya2

README.md
```diff
@@ -74,6 +74,31 @@ Notifications can be listed using the `listNotifications` method, which will ret
...
+// find the role that is allowed to read and write the shared file or folder 
+for ($roles as $role) {
+    if ($role->getDisplayName() === 'Editor') {
+        $editorRole = $role;
+        break;
+    }
+}
+
+// find all users with a specific surname
+$users = $ocis->listUsers("gurung");
```
amrita-shrestha, line 96: IMO
```suggestion
$users = $ocis->getUsers("gurung");
```

src/OcisResource.php
```diff
@@ -55,6 +56,24 @@ private function getMetadata(ResourceMetadata $property): array|string
         return $metadata[$property->getKey()];
     }
 
+    /**
+     * gets all possible permissions for the resource
```
phil-davis, line 60:
```suggestion
     * gets all possible roles for the resource
```

Is this intended to return all the roles that the current user has for the resource? (including any "sub-roles" that have a subset of the user's permissions.)
And the assumption is that the current user will be able to "pass on" (invite) to another user any of the returned roles when they "invite" (share) to some other user.

README.md
```diff
@@ -74,6 +74,31 @@ Notifications can be listed using the `listNotifications` method, which will ret
...
+```php
+// get the resources of a subfolder inside a drive
+$resources = $drive->listResources("/documents");
+
+// get all roles that are possible for that particular resource
+$roles = $resources[0]->getRoles();
+
+// find the role that is allowed to read and write the shared file or folder 
+for ($roles as $role) {
+    if ($role->getDisplayName() === 'Editor') {
```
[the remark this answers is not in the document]
individual-it: In real-live the roles will be displayed to the user, same as done in the ocis webUI, the user selects a role and then the backend creates a share with that role.
It's also possible to set detailed actions, see "send invite to user granting custom actions" in https://owncloud.dev/libre-graph-api/#/drives.permissions/Invite but I don't think we should implement that right now

src/OcisResource.php
individual-it: This function will get all (sharing)roles that a specific user can set for this resource. 
`"@libre.graph.permissions.roles.allowedValues"` part of https://owncloud.dev/libre-graph-api/#/drives.permissions/ListPermissions

src/OcisResource.php
```diff
@@ -73,6 +117,104 @@ private function getMetadata(ResourceMetadata $property): array|string
...
+            Ocis::createGuzzleConfig($this->connectionConfig, $this->accessToken)
+        );
+
+        if (array_key_exists('drivesPermissionsApi', $this->connectionConfig)) {
+            $apiInstance = $this->connectionConfig['drivesPermissionsApi'];
+        } else {
+            $apiInstance = new DrivesPermissionsApi(
+                $guzzle,
+                $this->graphApiConfig
+            );
```
individual-it, line 142: in other places we do this kind of dependency injection for unit tests through special setters, e.g. https://github.com/owncloud/ocis-php-sdk/blob/main/src/Ocis.php#L173-L179
If we are happy to use `connectionConfig` for that, we should refactor the other injections accordingly (in a different PR)

tests/unit/Owncloud/OcisPhpSdk/ResourceInviteTest.php
```diff
@@ -0,0 +1,162 @@
...
+                )
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider inviteDataProvider
+     * @param array<int, User|Group> $recipients
+     */
+    public function testInvite($recipients, ?\DateTime $expiration, DriveItemInvite $expectedInviteData): void
```
individual-it, line 128: This unit test is semi-useful, but at least it does some basic checks. As soon as the underlying ocis has the functionality implemented we should add integration tests

src/SharingRole.php
```diff
@@ -0,0 +1,72 @@
...
+        return $this->description;
+    }
+
+    /**
+     * @return int
+     */
+    public function getWeight(): int
+    {
+        return $this->weight;
+    }
```
amrita-shrestha, line 69: just curious what is use of weight?

amrita-shrestha (approved):
lgtm

src/SharingRole.php
individual-it: this is to help a client to sort the roles, e.g. `viewer` might have to be shown before `editor`, so sorting alphabetically does not make sense, so a client can sort by `weight`

PR #66, opened 2023-11-08
remove version number from service URL

individual-it:
after https://github.com/owncloud/libre-graph-api/pull/129 the version number needs to be removed also in our URL definitions


phil-davis (approved):

phil-davis:
Getting random fails like https://drone.owncloud.com/owncloud/ocis-php-sdk/320/3/2
```
Failed to clone the git@github.com:owncloud/libre-graph-api-php.git repository, try running in interactive mode so that you can enter your GitHub credentials
```

Happens in different pipelines each time I restart. I think that all the pipelines will pass, so we could just merge this.

amrita-shrestha:
> what if we have a fixed version of libre-graph-api and update it on some time interval?
> 
> ```json
> "require": {
>         ...
>         "owncloud/libre-graph-api-php": "dev-main#9b9dca2837c70797a198f21ddc2344046b34a809",
>         ...
>     },
> ```

this way our CI will not break every time. Until owncloud public libre-graph-api-php package. But we have to remeber to regular update commit id @individual-it @phil-davis

individual-it:
@amrita-shrestha good idea, let's do that in an other commit. Anyway we should publish both libs to packagist

PR #69, opened 2023-11-13
cache composer lock file  

amrita-shrestha:
This PR cache composer.lock file also


.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor/
+composer.lock
```
saw-jan, line 2: it is a good practice to commit the lock file as well.

amrita-shrestha: yeah but we have to locally update composer.lock file every time composer.json file changes

individual-it (approved):

.gitignore
saw-jan: >  we have to locally update composer.lock file every time we need an updated version

how is the update done? Won't the `composer` command update the lock file as well?

PR #79, opened 2023-11-17
Delete group

S-Panta:
This PR implements delete group functionality

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -8,6 +8,7 @@ use GuzzleHttp\Exception\GuzzleException;
 use OpenAPI\Client\Api\DrivesApi;
 use OpenAPI\Client\Api\DrivesGetDrivesApi;
 use OpenAPI\Client\Api\DrivesPermissionsApi;
+use OpenAPI\Client\Api\GroupApi;
 use OpenAPI\Client\Api\MeDrivesApi;
 use OpenAPI\Client\Api\UsersApi;
 use OpenAPI\Client\ApiException;
@@ -22,10 +23,10 @@ use Owncloud\OcisPhpSdk\Exception\HttpException;
 use Owncloud\OcisPhpSdk\Exception\NotFoundException;
 use Owncloud\OcisPhpSdk\Exception\UnauthorizedException;
 use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
+use Owncloud\OcisPhpSdk\Group;
 use Sabre\HTTP\ResponseInterface;
 use stdClass;
 use OpenAPI\Client\Api\GroupsApi;
-use Owncloud\OcisPhpSdk\Group;
 use OpenAPI\Client\Model\Group as OpenAPIGrop;
 
 /**
@@ -788,6 +789,17 @@ class Ocis
             throw new InvalidResponseException(
                 "createGroup returned an OdataError - " . $newlyCreatedGroup->getError()
             );
+        }
+    }
+
+    public function deleteGroup($groupName)
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        $groups = $this->getGroups($groupName);
+        // look $groups return value
+        foreach ($groups as $group) {
+            $groupId = $group->getId();
+            $apiInstance->deleteGroup($groupId);
         }
     }
 
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -782,6 +782,7 @@ class Ocis
         $group = new OpenAPIGrop(["display_name" => $groupName, "description" => $description]);
         try {
             $newlyCreatedGroup = $apiInstance->createGroup($group);
+
         } catch (ApiException $e) {
             throw ExceptionHelper::getHttpErrorException($e);
         }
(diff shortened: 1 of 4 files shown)


phil-davis:
@S-Panta  please rebase this so it "sees" the current master license.

amrita-shrestha (changes requested):

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -164,4 +165,5 @@ public function testGetGroupSort(OrderDirection $orderDirection, string $searchT
             $this->assertSame($resultGroups[$i], $groups[$i]->getDisplayName());
         }
     }
+
```
amrita-shrestha, line 168:
```suggestion
```

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -96,9 +94,9 @@ public function testGetGroupsExpanded()
     {
         $token = $this->getAccessToken('admin', 'admin');
         $ocis = new Ocis($this->ocisUrl, $token, ['verify' => false]);
-        $ocis->createGroup("violin-haters", "sss");
+        $ocis->createGroup("physics-lovers", "sss");
```
amrita-shrestha, line 97: write some meaning ful description
```suggestion
        $ocis->createGroup("physics-lovers", "physics lover group");
```

src/Ocis.php
```diff
@@ -770,12 +771,13 @@ public function getNotifications(): array
...
     * @throws HttpException
     */
-    public function createGroup(string $groupName, string $description = "")
+    public function createGroup(string $groupName, string $description = ""): void
     {
         $apiInstance = new GroupsApi($this->guzzle, $this->graphApiConfig);
         $group = new OpenAPIGrop(["display_name" => $groupName, "description" => $description]);
         try {
             $newlyCreatedGroup = $apiInstance->createGroup($group);
+
```
amrita-shrestha, line 780:
```suggestion
```

src/Ocis.php
```diff
@@ -786,4 +788,26 @@ public function createGroup(string $groupName, string $description = "")
...
+     * @throws HttpException
+     */
+    public function deleteGroup(Group $group): void
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        try {
+            $apiInstance->deleteGroup($group->getId());
+        } catch (ApiException $e) {
+            throw ExceptionHelper::getHttpErrorException($e);
+        }
```
amrita-shrestha, line 810: any reason for not handling ODataError?

S-Panta: delete group returns void as it's return type or it throws an exception. It doesn't throw any object and thus can't be checked whether the return value is of `OdataError instance`

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -44,6 +44,9 @@ public function tearDown(): void
             $drive->disable();
             $drive->delete();
         }
+        foreach($ocis->getGroups() as $group) {
+            $ocis->deleteGroup($group);
+        }
```
individual-it, line 49: I think we should remember the groups we have created in the tests and only delete those, not all. Just in case s.o. runs the test on a system that has already groups

individual-it (changes requested):
please also create at least a test that tries to delete a group but has no permission to do so

src/Ocis.php
amrita-shrestha: i can see ODataError inside libregraph deletegroup function. Please be sure what is actual behavior

S-Panta: I can see that. But `Odata error ` isn't caught as an Exception in our code. As far as I know, once the API Exception is caught, it can't test whether the variable is of  `OdataError` instances as the function return value is void only (though the exception inside libregraph returns OdataError).

individual-it: @S-Panta is right, `deleteGroup` does not return anything like other functions, so there is no odata error

individual-it:
looking at this code again I'm thinking if it would be better to move the delete function into the group class. So we just have to call `$group->delete()`
We can have a second function in `Ocis` called `deleteGroupById` that will get a group id and work from there. But now we anyway need an group object, but cannot delete the group in OOP manner.

@amrita-shrestha @S-Panta @nirajacharya2 thoughts?

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -21,6 +21,11 @@ class OcisPhpSdkTestCase extends TestCase
      */
     protected $createdDrives = [];
 
+    /**
+     * @var array <\Owncloud\OcisPhpSdk\Group>
+     */
+    protected $createdGroup = [];
```
individual-it, line 27: its an array, and obviously there cab be more that one group, so:
```suggestion
    protected $createdGroups = [];
```

individual-it (changes requested):

individual-it (approved):

individual-it (approved):

individual-it:
@amrita-shrestha please review

amrita-shrestha (approved):
lgtm

PR #80, opened 2023-11-18
move to libregraph generated with php-nextgen

individual-it:
some changes needed to use the new generator

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -27,12 +27,12 @@ class Group
         throw new InvalidResponseException(
             "Invalid id returned for group '" . print_r($openApiGroup->getId(), true) . "'"
         )
-        : $openApiGroup->getId();
+        : (string)$openApiGroup->getId();
         $this->displayName = empty($openApiGroup->getDisplayName()) ?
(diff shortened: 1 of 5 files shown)


src/Group.php
```diff
@@ -27,12 +27,12 @@ public function __construct(OpenApiGroup $openApiGroup)
         throw new InvalidResponseException(
             "Invalid id returned for group '" . print_r($openApiGroup->getId(), true) . "'"
         )
-        : $openApiGroup->getId();
+        : (string)$openApiGroup->getId();
```
individual-it, line 30: cast to string, even at this point it cannot be anything else, but it does confuse static analyser

src/Ocis.php
```diff
@@ -459,7 +460,7 @@ public function createDrive(
             [
                 'description' => $description,
                 'name' => $name,
-                'quota' => ['total' => $quota]
+                'quota' => new Quota(['total' => $quota])
```
individual-it, line 463: the type of the quota has to be `Quota`
this was a real bug

phil-davis (approved):

PR #84, opened 2023-11-20
get private link

individual-it:
Functionality to get the private link for a resource.
The link is delivered using the WebDAV properties, see https://owncloud.dev/apis/http/webdav/#listing-properties
I don't believe this is a very expensive data to generate, so I've decided to get it directly with the main request, alternatively we could send a new request just when calling `getPrivatelink`, but I don't think its worth it.

### DIFF
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -290,7 +290,25 @@ class Drive
         $resources = [];
         $webDavClient = $this->createWebDavClient();
         try {
-            $responses = $webDavClient->propFind(rawurlencode(ltrim($path, "/")), [], 1);
+            $properties = [
+                '{http://owncloud.org/ns}id',
+                '{http://owncloud.org/ns}fileid',
+                '{http://owncloud.org/ns}spaceid',
+                '{http://owncloud.org/ns}file-parent',
+                '{http://owncloud.org/ns}name',
+                '{DAV:}getetag',
+                '{http://owncloud.org/ns}permissions',
+                '{DAV:}resourcetype',
+                '{http://owncloud.org/ns}size',
+                '{DAV:}getcontentlength',
+                '{DAV:}getlastmodified',
+                '{http://owncloud.org/ns}tags',
+                '{http://owncloud.org/ns}favorite',
+                '{http://owncloud.org/ns}privatelink',
+                '{DAV:}getcontenttype',
+                '{http://owncloud.org/ns}checksums'
+            ];
+            $responses = $webDavClient->propFind(rawurlencode(ltrim($path, "/")), $properties, 1);
             foreach ($responses as $response) {
                 $resources[] = new OcisResource(
                     $response,
@@ -305,7 +323,8 @@ class Drive
             throw ExceptionHelper::getHttpErrorException($e);
         }
 
-        return $resources;
+        // make sure there is again an element with index 0
+        return array_values($resources);
     }
 
     /**
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -290,24 +290,10 @@ class Drive
         $resources = [];
         $webDavClient = $this->createWebDavClient();
         try {
-            $properties = [
-                '{http://owncloud.org/ns}id',
-                '{http://owncloud.org/ns}fileid',
-                '{http://owncloud.org/ns}spaceid',
-                '{http://owncloud.org/ns}file-parent',
-                '{http://owncloud.org/ns}name',
-                '{DAV:}getetag',
-                '{http://owncloud.org/ns}permissions',
-                '{DAV:}resourcetype',
-                '{http://owncloud.org/ns}size',
-                '{DAV:}getcontentlength',
-                '{DAV:}getlastmodified',
-                '{http://owncloud.org/ns}tags',
-                '{http://owncloud.org/ns}favorite',
-                '{http://owncloud.org/ns}privatelink',
-                '{DAV:}getcontenttype',
-                '{http://owncloud.org/ns}checksums'
-            ];
+            $properties = [];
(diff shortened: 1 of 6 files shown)


src/Drive.php
```diff
@@ -290,7 +290,25 @@ public function getResources(string $path = "/"): array
...
+                '{DAV:}resourcetype',
+                '{http://owncloud.org/ns}size',
+                '{DAV:}getcontentlength',
+                '{DAV:}getlastmodified',
+                '{http://owncloud.org/ns}tags',
+                '{http://owncloud.org/ns}favorite',
+                '{http://owncloud.org/ns}privatelink',
+                '{DAV:}getcontenttype',
+                '{http://owncloud.org/ns}checksums'
+            ];
```
amrita-shrestha, line 310: do we need to send all properties?

amrita-shrestha (commented):
other lgtm

src/Drive.php
individual-it: yes, those are the properties that we currently support, all except `'{http://owncloud.org/ns}privatelink'` come by default with a PROPFIND without requesting a property, so if we want to keep backward compatibility and still support all the properties, they need all to be listed here

src/Drive.php
```diff
@@ -304,7 +322,8 @@ public function getResources(string $path = "/"): array
             throw ExceptionHelper::getHttpErrorException($e);
         }
 
-        return $resources;
+        // make sure there is again an element with index 0
```
phil-davis, line 325: should we also put a comment for line 320 (unset) to mention why that is done - I think that is removing the entry for a folder itself.

src/OcisResource.php
```diff
@@ -402,4 +388,20 @@ public function getCheckSums(): array
...
+     * Returns the private link to the resource.
+     * This link can be used by any user with the correct permissions to navigate to the resource in the web UI.
+     * The link is urldecoded.
+     * @return string
+     * @throws InvalidResponseException
+     */
+    public function getPrivatelink(): string
+    {
+        $privateLink = $this->getMetadata(ResourceMetadata::PRIVATELINK);
+        if (is_array($privateLink)) {
```
phil-davis, line 402: How does this happen - I guess that if the resource does not have any private link, then there is an empty array at this point? Maybe add a comment?

src/Drive.php
```diff
@@ -290,7 +290,25 @@ public function getResources(string $path = "/"): array
         $resources = [];
         $webDavClient = $this->createWebDavClient();
         try {
-            $responses = $webDavClient->propFind(rawurlencode(ltrim($path, "/")), [], 1);
+            $properties = [
```
phil-davis, line 293: Can we construct these property strings by using the names in the ResourceMetadata enum?

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -0,0 +1,98 @@
...
+            );
+            $this->assertMatchesRegularExpression(
+                "?^" . $this->ocisUrl . '/f/'. $this->getFileIdRegex() . "$?i",
+                $resource->getPrivatelink()
+            );
+            try {
+                $this->assertEquals('file', $resource->getType());
+            } catch (\Exception $e) {
+                $this->assertEquals('folder', $resource->getType());
+            }
```
phil-davis, line 58: phpunit has `logicalOr` - maybe this can be something like:
```
$this-logicalOr($this->assertEquals('file', $resource->getType()), $this->assertEquals('folder', $resource->getType()));
```
and similar for some other try-catch here.

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -2,24 +2,31 @@
 
 namespace integration\Owncloud\OcisPhpSdk;
 
-require_once __DIR__ . '/OcisPhpSdkTestCase.php';
-
 use GuzzleHttp\Client;
+use Owncloud\OcisPhpSdk\Drive; // @phan-suppress-current-line PhanUnreferencedUseNormal it's used in a comment
```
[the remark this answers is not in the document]
individual-it: this is just nice to help the IDE to guess correctly the type and then help with auto-completion. If that is annoying for others we can get rid of it

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -44,6 +51,17 @@ public function tearDown(): void
             $drive->disable();
             $drive->delete();
         }
+        foreach ($this->createdResources as $resource) {
+            /**
+             * @var Drive $personalDrive
+             */
```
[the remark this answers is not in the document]
individual-it: this is just nice to help the IDE to guess correctly the type and then help with auto-completion. If that is annoying for others we can get rid of it

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
individual-it: done

src/Drive.php
individual-it: done

src/OcisResource.php
individual-it: this really should not happen, but the analyzer don't know, so I've made it more explicit by throwing an exception

src/Drive.php
individual-it: done

amrita-shrestha (approved):
lgtm but need to resolve conflict

PR #87, opened 2023-11-21
use driveId instead or itemId as first argument in API calls

individual-it:
in these API calls we actually need the `driveId` not the `itemId` as the first argument see https://owncloud.dev/libre-graph-api/#/drives.permissions/Invite

So forwarding the driveId from the drive class down to the resource and using it there


individual-it:
CC @fschade

phil-davis (approved):
Looks reasonable.

PR #90, opened 2023-11-22
allow a share to be updated

individual-it:
1. `invite()` function returns an array of `ShareCreated` objects
2. `ShareCreated` & `ShareLink` extend the `Share` class, which combines functions/properties that are needed in both places
3. the role in `ShareCreated` can be changed (and of course anything that is extended from `Share`)

### DIFF
--- a/src/Share.php
+++ b/src/Share.php
@@ -105,7 +105,7 @@ class Share
     }
 
     /**
-     * permanently delete the current share or share link
+     * Permanently delete the current share or share link
      *
      * @throws BadRequestException
      * @throws ForbiddenException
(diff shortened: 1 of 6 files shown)


src/Share.php
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * @throws UnauthorizedException
+     * @throws ForbiddenException
+     * @throws InvalidResponseException
+     * @throws BadRequestException
+     * @throws HttpException
+     * @throws NotFoundException
+     */
+    public function setExpiration(\DateTime $expiration): bool
```
amrita-shrestha, line 139: we can set role and expiration for share and sharelink, why setexpiration and setrole in different class?

individual-it: I've improved the comments of some functions, hope that explains it

amrita-shrestha (approved):
lgtm

PR #91, opened 2023-11-23
create OcisResource object from file-id

individual-it:
- If only a file-id is known, allow the user to create an `OcisResource` just using that. The use-case is that an application might have stored the file-id in the database and then wants to use all the abilities of `OcisResource` with it.
- move the functions to get the content of a file to the `OcisResource` class. Does that make sense? We can also keep both, but the naming might be confusing.


amrita-shrestha (approved):
lgtm

phil-davis (approved):
LGTM

Issue #93, opened 2023-11-27
Add delete user from group


amrita-shrestha:
In sdk, group can be created. Group class contains a method to delete and addUser in it. We can add a method to make API request which delete users from group

PR #95, opened 2023-11-27
remove user from group function implementation

S-Panta:
This PR implements user removal form Group. 
Related issues: https://github.com/owncloud/ocis-php-sdk/issues/93

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -161,8 +161,8 @@ class Group
 
     /**
      * remove an existing user from group
-     * @param User $user
      *
+     * @param User $user
      * @return void
      * @throws BadRequestException
      * @throws ForbiddenException
@@ -179,6 +179,7 @@ class Group
         } catch (ApiException $e) {
             throw ExceptionHelper::getHttpErrorException($e);
         }
+        $this->members[] = $user;
     }
 
     /**
--- a/src/Group.php
+++ b/src/Group.php
@@ -159,6 +159,16 @@ class Group
         $this->members[] = $user;
     }
 
+    public function removeUser($user): void
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        try {
+            $apiInstance->deleteMember($this->getId(), $user->getId());
+        } catch (ApiException $e) {
+            throw ExceptionHelper::getHttpErrorException($e);
+        }
+    }
+
     /**
      * delete an existing group (if the user has the permission to do so)
      *
--- a/src/Group.php
+++ b/src/Group.php
@@ -159,6 +159,18 @@ class Group
         $this->members[] = $user;
     }
 
+    /**
+     * remove an existing user from group
+     * @param User $user
+     *
+     * @return void
+     * @throws BadRequestException
+     * @throws ForbiddenException
+     * @throws NotFoundException
+     * @throws UnauthorizedException
+     * @throws \InvalidArgumentException
+     * @throws HttpException
+     */
     public function removeUser($user): void
     {
         $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
--- a/src/Group.php
+++ b/src/Group.php
@@ -160,7 +160,7 @@ class Group
     }
 
     /**
-     * remove an existing user from group
+     * Remove an existing user from the current group
      *
      * @param User $user
      * @return void
@@ -180,7 +180,7 @@ class Group
             throw ExceptionHelper::getHttpErrorException($e);
         }
         foreach($this->members as $memberIndex => $member) {
-            if ($member->getDisplayName() === $user->getDisplayName()) {
+            if ($member->getId() === $user->getId()) {
                 unset($this->members[$memberIndex]);
             }
         }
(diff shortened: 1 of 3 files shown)


src/Group.php
```diff
@@ -157,6 +157,28 @@ public function addUser($user): void
...
+     * @throws HttpException
+     */
+    public function removeUser($user): void
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        try {
+            $apiInstance->deleteMember($this->getId(), $user->getId());
+        } catch (ApiException $e) {
+            throw ExceptionHelper::getHttpErrorException($e);
+        }
```
amrita-shrestha, line 179: don't we have to update $this->members[] after removing user from group?

src/Group.php
```diff
@@ -159,6 +159,34 @@ public function addUser($user): void
         $this->members[] = $user;
     }
 
+    /**
+     * remove an existing user from group
```
individual-it, line 163:
```suggestion
     * Remove an existing user from the current group
```

src/Group.php
```diff
@@ -159,6 +159,34 @@ public function addUser($user): void
...
+    public function removeUser($user): void
+    {
+        $apiInstance = new GroupApi($this->guzzle, $this->graphApiConfig);
+        try {
+            $apiInstance->deleteMember($this->getId(), $user->getId());
+        } catch (ApiException $e) {
+            throw ExceptionHelper::getHttpErrorException($e);
+        }
+        foreach($this->members as $memberIndex => $member) {
+            if ($member->getDisplayName() === $user->getDisplayName()) {
```
individual-it, line 183: can we compare by id and not display name? The display name might not be unique

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -127,6 +127,36 @@ public function testAddUserToGroup(): void
...
+     * @return void
+     */
+    public function testRemoveExistingUserFromGroup(): void
+    {
+        $token = $this->getAccessToken('admin', 'admin');
+        $ocis = new Ocis($this->ocisUrl, $token, ['verify' => false]);
+        $users = $ocis->getUsers('admin');
+        $philosophyHatersGroup =  $ocis->createGroup("philosophy-haters", "philosophy haters group");
+        $this->createdGroups = [$philosophyHatersGroup];
+        $philosophyHatersGroup->addUser($users[0]);
```
individual-it, line 140: I think it would be good to have more than one user in the group, just to make sure the `removeUser` function not removing all users from the group, but the correct one only

individual-it (changes requested):

S-Panta:
A finding that may help. While implementing `addUser`, we store the `user` object in our PHP code in `this->members` and don't get cleanup in the teardown. Although it doesn't hamper our test case, is this a good practice?
@amrita-shrestha  @individual-it @nirajacharya2

amrita-shrestha:
> A finding that may help. While implementing `addUser`, we store the `user` object in our PHP code in `this->members` and don't get cleanup in the teardown. Although it doesn't hamper our test case, is this a good practice? @amrita-shrestha @individual-it @nirajacharya2

cleanup is best practice @S-Panta

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -131,6 +131,48 @@ public function testAddUserToGroup(): void
...
+            $philosophyHatersGroup->addUser($user);
+        }
+        $initialMemberCount = count($philosophyHatersGroup->getMembers());
+        foreach ($users as $user) {
+            if($user->getDisplayName() === "Albert Einstein") {
+                $philosophyHatersGroup->removeUser($user);
+            }
+        }
+        $adminUserName = $users[0]->getDisplayName();
+        foreach ($ocis->getGroups(expandMembers: true) as $group) {
```
individual-it, line 156: there should be only one group, and if there were multiple groups the whole test would fail, so why having the `foreach`?

individual-it (changes requested):
maybe also add a test where you delete a user with an non-admin user

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
S-Panta: I think the test coverage can be extended only if the issue regarding `getGroup` from another user is done.

individual-it:
@S-Panta you still can write a test without getting a user as a normal user.
- as admin get a user 
- as admin create a group
- as admin add the user to the group
- as user use the own user object trying to remove yourself from the group

amrita-shrestha:
> @S-Panta you still can write a test without getting a user as a normal user.
> 
> * as admin get a user
> * as admin create a group
> * as admin add the user to the group
> * as user use the own user object trying to remove yourself from the group

@individual-it @S-Panta please correct if i am wrong . Problem is that normal user can't get group id too

S-Panta:
Also, deleteuserfromgroup takes only 

> > @S-Panta you still can write a test without getting a user as a normal user.
> > 
> > * as admin get a user
> > * as admin create a group
> > * as admin add the user to the group
> > * as user use the own user object trying to remove yourself from the group
> 
> @individual-it @S-Panta please correct if i am wrong . Problem is that normal user can't get group id too

Yes, AFAIK only admin can get group id.

PR #98, opened 2023-11-28
allow removal of expiration date of shares

individual-it:
after the fix in the [libre-graph-api](https://github.com/owncloud/libre-graph-api/pull/145) we can now delete the expiration date

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "^8.1",
         "sabre/dav": "^4.4",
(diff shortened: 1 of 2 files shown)


phil-davis (approved):
Do we have test coverage of this yet?
Can we add a test case for removing the expiration date?

individual-it:
@phil-davis we can, as soon as ocis has the functionality implemented ;-)
Currently its all only in the api definition.

PR #100, opened 2023-11-29
add mountpoit drivetype

individual-it:
received shares are listed as 'mountpoint' drivetype when getting my drives (`/v1.0/me/drives`), so add this type to the list


amrita-shrestha (approved):
lgtm

PR #101, opened 2023-11-30
ability to check if drive is disabled

individual-it:
When listing drives (spaces) it might be handy to know if a particular drive is currently disabled to be able to display it differently or not at all.

### DIFF
--- a/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace integration\Owncloud\OcisPhpSdk;
+
+require_once __DIR__ . '/OcisPhpSdkTestCase.php';
+
+class DriveTest extends OcisPhpSdkTestCase
+{
+    public function testDisableDrive(): void
+    {
+        $ocis = $this->getOcis('admin', 'admin');
+        $drive = $ocis->createDrive('test drive');
+        $this->createdDrives[] = $drive->getId();
+        $this->assertFalse($drive->isDisabled());
+        $drive->disable();
+        $this->assertTrue($drive->isDisabled());
+    }
+}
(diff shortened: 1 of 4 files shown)


tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -112,6 +112,12 @@ protected function getAccessToken(string $username, string $password): string
         return $accessTokenResponse['access_token'];
     }
 
+    protected function getOcis(string $username, string $password): Ocis
```
amrita-shrestha, line 115: i think inside this file there are some place where we can use this function?
```
        $token = $this->getAccessToken($username, $password);
        return new Ocis($this->ocisUrl, $token, ['verify' => false]);
```

individual-it: Good point, thank you. I would suggest to do that in a different PR. I've created an issue, so that we don't for get it #104

phil-davis (approved):

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -0,0 +1,18 @@
...
+
+class DriveTest extends OcisPhpSdkTestCase
+{
+    public function testDisableDrive(): void
+    {
+        $ocis = $this->getOcis('admin', 'admin');
+        $drive = $ocis->createDrive('test drive');
+        $this->createdDrives[] = $drive->getId();
+        $this->assertFalse($drive->isDisabled());
+        $drive->disable();
```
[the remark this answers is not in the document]
individual-it: Yes, we anyway have a problem with that, e.g. with Quotas.
One reads a drive and checks the quota, then uploads files but the quota in the object stays the same. We need to think about which getters need to read data from the server.
I will create an issue

PR #106, opened 2023-12-01
get receiver of share

individual-it:
If one got a `ShareCreated` object its handy to be able to find out who the receiver of the share is. For a share just created it might not be so useful as the receiver is known, but when finding shares by `SharedByMe` its useful.

### DIFF
--- a/src/ShareCreated.php
+++ b/src/ShareCreated.php
@@ -55,4 +55,37 @@ class ShareCreated extends Share
         $this->apiPermission = $apiPermission;
         return true;
     }
+
+    /**
+     * @throws UnauthorizedException
+     * @throws ForbiddenException
+     * @throws InvalidResponseException
+     * @throws HttpException
+     * @throws BadRequestException
+     * @throws NotFoundException
+     */
+    public function getReceiver(): User|Group
+    {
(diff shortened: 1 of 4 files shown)


src/ShareCreated.php
```diff
@@ -55,4 +55,37 @@ public function setRole(SharingRole $role): bool
...
+     * @throws UnauthorizedException
+     * @throws ForbiddenException
+     * @throws InvalidResponseException
+     * @throws HttpException
+     * @throws BadRequestException
+     * @throws NotFoundException
+     */
+    public function getReceiver(): User|Group
+    {
+        $ocis = new Ocis($this->serviceUrl, $this->accessToken, $this->connectionConfig);
```
individual-it, line 69: is this a legitimate approach?
I wanted to have `getUserById` and `getGroupById` in the `Ocis`class, but also feels strange to create an `Ocis` object here.
Alternatively I could have `getUserById` and `getGroupById` in some other helper class.

individual-it: yes, but which one and how to keep it so, that its easy to use for the developers?

amrita-shrestha (approved):
lgtm

Issue #83, opened 2023-11-20
unify dependency injection for unit tests


individual-it:
Make all dependency injections of  `OpenAPI\Client\Api\*` work the same as: https://github.com/owncloud/ocis-php-sdk/pull/61#pullrequestreview-1731349780

individual-it:
fixed by #108

PR #108, opened 2023-12-04
dependency injector made consistent

S-Panta:
This PR unify the dependency injector used for unit testing and remove the setter function and use configuration array to mock different drive config.
Related issue:https://github.com/owncloud/ocis-php-sdk/issues/83


individual-it (approved):
nice!
fixes #83

PR #114, opened 2023-12-05
delete accept & decline functions

### LINKED ISSUE #92: Add accept and decline share
Currently, there is no API endpoint provided in https://owncloud.dev/libre-graph-api/
After api endpoint updated in swagger, add accept and decline share method in sdk

individual-it:
these functions are irrelevant for now

fixes #92


amrita-shrestha (commented):
so we are not considering the environment variable which disables auto-accept false. Currently, ocis has an environment variable for that. If so then we can close this issue. @individual-it

individual-it:
the env. variable does not matter, we currently don't need this functionality in the SDK

amrita-shrestha (approved):
lgtm

Issue #110, opened 2023-12-04
unify handling cache invalidation


individual-it:
in some cases we can have invalid data in the objects, e.g.
1. list drives
2. upload files to one drive
3. check quota of the drive

the quota will returning the data before the file upload.
What do we want to do in those cases? Recheck the server status in every `get` function?

Thought? @amrita-shrestha @phil-davis @nirajacharya2 @S-Panta

amrita-shrestha:
@individual-it  doesn't this issue is fixed by phil ji in sabre-dav

individual-it:
No, this is totally unrelated.

individual-it:
This will be not implemented in the first release, but only in case the cache-invalidation causes problems

PR #116, opened 2023-12-05
test deleting drive

individual-it:
on top of #115
test delete drive

### DIFF
--- a/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
@@ -3,18 +3,21 @@
 namespace integration\Owncloud\OcisPhpSdk;
 
 use Owncloud\OcisPhpSdk\Drive;
+use Owncloud\OcisPhpSdk\Exception\BadRequestException;
 use Owncloud\OcisPhpSdk\Exception\NotFoundException;
+use Owncloud\OcisPhpSdk\Ocis;
 
 require_once __DIR__ . '/OcisPhpSdkTestCase.php';
 
 class DriveTest extends OcisPhpSdkTestCase
 {
     private Drive $drive;
+    private Ocis $ocis;
     public function setUp(): void
     {
         parent::setUp();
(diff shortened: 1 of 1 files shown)


tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -2,17 +2,70 @@
...
+        $this->assertFalse($this->drive->isDisabled());
+    }
+
+    public function testDeleteDrive(): void
+    {
+        $this->drive->disable();
+        $this->drive->delete();
+        $this->expectException(NotFoundException::class);
+        $this->expectExceptionMessage('itemNotFound - no drive returned from storage');
+        $this->ocis->getDriveById($this->drive->getId());
```
amrita-shrestha, line 53: don't we have to remove id from $this->createdDrives[] after delete drive

individual-it: we do not update the object itself

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -2,17 +2,70 @@
...
+        $this->expectException(NotFoundException::class);
+        $this->expectExceptionMessage('itemNotFound - drive not found');
+        $this->drive->delete();
+    }
+
+    public function testDeleteEnabledDrive(): void
+    {
+        $this->expectException(BadRequestException::class);
+        $this->expectExceptionMessage('invalidRequest - error: bad request: can\'t purge enabled space');
+        $this->drive->delete();
```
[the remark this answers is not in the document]
individual-it: It will never come to the assertion, because `delete` will throw an exception

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
individual-it: That is a more general question: #110

amrita-shrestha (approved):
lgtm

PR #118, opened 2023-12-08
Add test for move 

amrita-shrestha:
This add tests for move and upload


individual-it (changes requested):
1. please move the Trashbin things to an other PR, that is not related to move
2. I would suggest to have an extra function to get resources of Trashbin, there will be more things that are different to normal `getResources`, e.g. we need other properties. This is what web sends

```
<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
  <d:prop>
    <d:getcontentlength/>
    <d:resourcetype/>
    <oc:trashbin-original-location/>
    <oc:trashbin-original-filename/>
    <oc:trashbin-delete-datetime/>
    <oc:permissions/>
    <oc:file-parent/>
  </d:prop>
</d:propfind>
```

individual-it (approved):

PR #119, opened 2023-12-08
support proxy settings

individual-it:
allow to set proxy settings to force all connections over the proxy.

How to test:
1. install mitmproxy.org
2. set proxy settings, which are the same format like in guzzle: https://docs.guzzlephp.org/en/stable/request-options.html#proxy
3. check if all expected requests go through mitmproxy


amrita-shrestha (approved):
lgtm

PR #122, opened 2023-12-13
adding test for listing users and groups by normal user

nirajacharya2:
tests for listing groups and users as a normal user

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -591,6 +591,7 @@ class Ocis
 
     /**
      * retrieve users known by the system
+     * NOTE: if this function is used by a normal user a search string with at least 3 characters should be provided
      *
      * @param string|null $search
      * @return array<User>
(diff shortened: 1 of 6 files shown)


src/Ocis.php
```diff
@@ -591,6 +591,7 @@ public function getResourceById(string $fileId): OcisResource
 
     /**
      * retrieve users known by the system
+     * NOTE: if this function is used by a normal user a search string with atleast 3 characters should be provided
```
individual-it, line 594:
```suggestion
     * NOTE: if this function is used by a normal user a search string with at least 3 characters should be provided
```

tests/integration/Owncloud/OcisPhpSdk/GroupsTest.php
```diff
@@ -138,4 +138,41 @@ public function testDeleteGroup(): void
...
+     * @return void
+     */
+    public function testDeleteGroupByIdNoPermission(): void
+    {
+        $token = $this->getAccessToken("admin", "admin");
+        $ocis = new Ocis($this->ocisUrl, $token, ["verify" => false]);
+        $philosophyHatersGroup = $ocis->createGroup("philosophy-haters", "philosophy haters group");
+        $this->createdGroups = [$philosophyHatersGroup];
+        $token = $this->getAccessToken('einstein', 'relativity');
+        $ocisEinestine = new Ocis($this->ocisUrl, $token, ["verify" => false]);
```
individual-it, line 172:
```suggestion
        $ocisEinstein = new Ocis($this->ocisUrl, $token, ["verify" => false]);
```

individual-it (changes requested):

individual-it (approved):

Issue #113, opened 2023-12-05
make getSharedByMe return the correct types of objects


individual-it:
`getSharedByMe` also gets created links, but it always returns `ShareCreated` objects. For links it should return `ShareLink` objects

individual-it:
fixed by #127

PR #127, opened 2023-12-15
modified getSharedbyMe function

S-Panta:
This PR add step to return sharelink object created by the user.
Related Issue:https://github.com/owncloud/ocis-php-sdk/issues/113

### DIFF
--- a/tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
@@ -4,6 +4,7 @@ namespace integration\Owncloud\OcisPhpSdk;
 
 require_once __DIR__ . '/OcisPhpSdkTestCase.php';
 
+use OpenAPI\Client\Model\SharingLinkType;
 use Owncloud\OcisPhpSdk\Drive;
 use Owncloud\OcisPhpSdk\DriveOrder;
 use Owncloud\OcisPhpSdk\DriveType;
@@ -12,6 +13,10 @@ use Owncloud\OcisPhpSdk\Group;
 use Owncloud\OcisPhpSdk\Ocis;
 use Owncloud\OcisPhpSdk\OrderDirection;
 use Owncloud\OcisPhpSdk\Exception\NotFoundException;
+use Owncloud\OcisPhpSdk\Exception\UnauthorizedException;
+use Owncloud\OcisPhpSdk\ShareCreated;
+use Owncloud\OcisPhpSdk\ShareLink;
+use Owncloud\OcisPhpSdk\ShareReceived;
 
 class OcisTest extends OcisPhpSdkTestCase
 {
@@ -452,6 +457,23 @@ class OcisTest extends OcisPhpSdkTestCase
         $this->createdGroups = [$physicsLoversGroup];
     }
 
+    /**
+     * @return void
+     */
+    public function testDeleteGroupByIdNoPermission(): void
+    {
+        $ocis = $this->getOcis('admin', 'admin');
+        $philosophyHatersGroup = $ocis->createGroup("philosophy-haters", "philosophy haters group");
+        //any user other than admin can't get Group ID because of bug, thus bypassing this step
+        //Todo : make Einstein get Group ID after this bug is solved.
+        $groupId = $philosophyHatersGroup->getId();
+        $this->createdGroups = [$philosophyHatersGroup];
+        $token = $this->getAccessToken('einstein', 'relativity');
+        $ocis = new Ocis($this->ocisUrl, $token, ["verify" => false]);
+        $this->expectException(UnauthorizedException::class);
+        $ocis->deleteGroupByID($groupId);
+    }
+
     private function getPersonalDrive(Ocis $ocis): Drive
     {
         return $ocis->getMyDrives(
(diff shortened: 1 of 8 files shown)


tests/integration/Owncloud/OcisPhpSdk/ResourceInviteTest.php
```diff
@@ -255,14 +256,17 @@ public function testGetReceiversOfShareReturnedBySharedByMe(): void
...
         for($i = 0; $i < 6; $i++) {
-            $this->assertThat(
-                $shares[$i]->getReceiver()->getDisplayName(),
-                $this->logicalOr(
-                    $this->equalTo("philosophy-haters"),
-                    $this->equalTo("Marie Curie"),
-                    $this->equalTo("Albert Einstein")
-                )
-            );
+            if($shares[$i] instanceof ShareCreated) {
```
individual-it, line 259: there should be only `ShareCreated` items here at this time, or?
The danger with this is, that if we would have a bug and return something else than `ShareCreated` there would be no assertions done.
To make the analyzer happy maybe it would be enough to check `assertContainsOnly`

individual-it (changes requested):
a simple test where there is one link-share and a normal share would be nice

tests/integration/Owncloud/OcisPhpSdk/ResourceInviteTest.php
phil-davis: There is https://docs.phpunit.de/en/9.6/assertions.html#assertinstanceof

So you could not put this `if` statement, but instead just have an `assertInstanceOf` before `assertThat` - that way the test will fail early if `$shares[$i]` is not an instance of `ShareCreated`

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+        foreach ($resources as $resource) {
+            if ($resource->getName() === 'newFolder') {
+                $sharedResource = $resource;
+            }
+        }
+        if(empty($sharedResource)) {
+            throw new \Error(
+                "resource not found "
+            );
+        }
```
individual-it, line 579: use assertion

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+                $editorRole = $role;
+            }
+        }
+        if(empty($editorRole)) {
+            throw new \Error(
+                "Editor role not found "
+            );
+        }
+        $sharedResource->invite([$einstein], $editorRole);
+        $myShare = $ocis->getSharedByMe();
```
individual-it, line 592:
```suggestion
        $myShares = $ocis->getSharedByMe();
```

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+        $this->assertInstanceOf(ShareCreated::class, $myShare[0]);
+        $this->assertEquals($myShare[0]->getReceiver()->getDisplayName(), 'Albert Einstein');
```
individual-it, line 594: maybe check also the resource name

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+                $editorRole = $role;
+            }
+        }
+        if(empty($editorRole)) {
+            throw new \Error(
+                "Editor role not found "
+            );
+        }
+        $sharedResource->invite([$einstein], $editorRole);
+        $sharedResource->createSharingLink(SharingLinkType::VIEW, new \DateTimeImmutable('2023-12-31 01:02:03.456789'), 'Test@123', '');
```
individual-it, line 628: long line

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+        $myShare = $ocis->getSharedByMe();
+        $this->assertInstanceOf(ShareCreated::class, $myShare[0]);
+        $this->assertInstanceOf(ShareLink::class, $myShare[1]);
```
individual-it, line 631: Is the order guaranteed ?

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+                "Editor role not found "
+            );
+        }
+        $sharedResource->invite([$einstein], $editorRole);
+        $myShare = $ocis->getSharedByMe();
+        $this->assertInstanceOf(ShareCreated::class, $myShare[0]);
+        $this->assertEquals($myShare[0]->getReceiver()->getDisplayName(), 'Albert Einstein');
+    }
+
+    public function testGetShareCreatedAndShareCreatedLinkByMe(): void
```
individual-it, line 597:
```suggestion
    public function testGetSharedByMeContainingShareCreatedAndShareLink(): void
```

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+            );
+        }
+        $sharedResource->invite([$einstein], $editorRole);
+        $sharedResource->createSharingLink(SharingLinkType::VIEW, new \DateTimeImmutable('2023-12-31 01:02:03.456789'), 'Test@123', '');
+        $myShare = $ocis->getSharedByMe();
+        $this->assertInstanceOf(ShareCreated::class, $myShare[0]);
+        $this->assertInstanceOf(ShareLink::class, $myShare[1]);
+    }
+
+    public function testGetSharedWithMe(): void
```
individual-it, line 634: How is that related to `getSharedByMe`? Should that be in a different PR?

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
         $ocis = $this->getOcis('admin', 'admin');
         $ocis->getResourceById('not-existing-id');
     }
+
+    public function testGetShareCreatedByMe(): void
```
individual-it, line 561:
```suggestion
    public function testGetSharedByMe(): void
```

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -452,6 +457,23 @@ public function testDeleteGroupById(): void
         $this->createdGroups = [$physicsLoversGroup];
     }
 
+    /**
+     * @return void
+     */
+    public function testDeleteGroupByIdNoPermission(): void
```
individual-it, line 463: how is that related to `getSharedByMe`? Should that be a separate PR?

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
```diff
@@ -535,4 +557,115 @@ public function testGetResourceInvalidId(): void
...
+        foreach($sharedResource->getRoles() as $role) {
+            if($role->getDisplayName() === 'Editor') {
+                $editorRole = $role;
+            }
+        }
+        if(empty($editorRole)) {
+            throw new \Error(
+                "Editor role not found "
+            );
+        }
```
individual-it, line 626: could be moved in `setUp()`

individual-it (changes requested):
please consider to move the tests into an other file `GetSharedByMeTest.php`

tests/integration/Owncloud/OcisPhpSdk/OcisTest.php
S-Panta: I've tested and found that the order remains `ShareCreated` and `ShareLink`.

individual-it (approved):

PR #129, opened 2023-12-18
adding test for getting roles of a resource

nirajacharya2:


individual-it (approved):

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -284,4 +285,30 @@ public function testUploadFileStream(): void
         $content = $this->getContentOfResource425Save($resources[0]);
         $this->assertEquals('some content', $content);
     }
+
+    public function testGetRoles(): void
```
amrita-shrestha, line 289: does file and folder has same resources

individual-it: Not necessary, but that is ocis behavior, we don't need to test that, we have to work with whatever ocis sends us

individual-it:
@nirajacharya2 please rebase and solve conflict

amrita-shrestha (approved):

PR #130, opened 2023-12-18
use spaceid from PROPFIND in OcisResource

individual-it:
with https://github.com/owncloud/ocis/pull/7975 we don't need to try to get the 'real' drive-id from the server, we can directly use the `spaceid` from the PROPFIND response as drive-id in the graph API.

Fixes: https://github.com/owncloud/moodle-repository_ocis/issues/49

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,3 +1,3 @@
 # The version of OCIS to use in pipelines that test against OCIS
-OCIS_COMMITID=55f3ef0aff4c2ee58e4cf7b8a38391b43c0c576f
-OCIS_BRANCH=master
+OCIS_COMMITID=0c68af5369aed5fe74a5952ed48a06e3cb335b17
+OCIS_BRANCH=fix-ocdav-spaceid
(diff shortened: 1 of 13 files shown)


tests/integration/compose.yaml
```diff
@@ -52,7 +52,7 @@ services:
 
 
   ocis:
-    image: owncloud/ocis:${OCIS_VERSION:-latest}
+    image: ocis:sharing
```
individual-it, line 55: to be reverted

.drone.env
```diff
@@ -1,3 +1,3 @@
 # The version of OCIS to use in pipelines that test against OCIS
-OCIS_COMMITID=55f3ef0aff4c2ee58e4cf7b8a38391b43c0c576f
-OCIS_BRANCH=master
+OCIS_COMMITID=0c68af5369aed5fe74a5952ed48a06e3cb335b17
+OCIS_BRANCH=fix-ocdav-spaceid
```
individual-it, line 3: to be reverted and updated after https://github.com/owncloud/ocis/pull/7975 is merged

amrita-shrestha (approved):
lgtm

PR #135, opened 2023-12-21
removing '-' from group names

nirajacharya2:
currently in ocis we can't search for groups if the search query contains '-'. so its better not create groups with that character in our tests


phil-davis:
@nirajacharya2 why are we doing this?
Please write something in the PR description.

individual-it (approved):

Issue #99, opened 2023-11-29
unify where data recieved from server is validated


individual-it:
When creating objects with data that ocis has sent us we validate it, but sometimes we validate it in the constructor, e.g.: https://github.com/owncloud/ocis-php-sdk/blob/main/src/Group.php#L66-L75
sometimes we validate it in the getters: https://github.com/owncloud/ocis-php-sdk/blob/main/src/ShareRecieved.php#L35-L42

We should decide what to do and do it the same everywhere

individual-it:
Let's have all the check in the corresponding getters and ONLY use getters ourself also in the functions of the same class

amrita-shrestha:
Todo:
-  Notification.php

amrita-shrestha:
@nirajacharya2 if nothing then close the issue

PR #157, opened 2023-12-28
refactoring user class to validate data from server inside getters

nirajacharya2:
previously we validated the response from server in constructor in user class but now this PR moves that logic to the getters.

### DIFF
--- a/tests/unit/Owncloud/OcisPhpSdk/UserTest.php
+++ b/tests/unit/Owncloud/OcisPhpSdk/UserTest.php
@@ -113,6 +113,9 @@ class UserTest extends TestCase
             ]
         );
         $user = new SdkUser($libUser);
+        $user->getDisplayName();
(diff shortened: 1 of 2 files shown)


tests/unit/Owncloud/OcisPhpSdk/UserTest.php
```diff
@@ -113,6 +113,9 @@ public function testInvalidUserData(array $data)
             ]
         );
         $user = new SdkUser($libUser);
+        $user->getDisplayName();
+        $user->getMail();
+        $user->getId();
```
individual-it, line 118: the code will never get here, because `getDisplayName` throws the exception. If you want to test every function you will have to have more test cases

individual-it (changes requested):

individual-it (approved):

PR #161, opened 2023-12-29
refactoring sharing role class to validate data from server inside getters

nirajacharya2:
previously we validated the response from server in constructor in sharing role class but now this PR moves that logic to the getters.

### DIFF
--- a/src/SharingRole.php
+++ b/src/SharingRole.php
@@ -11,10 +11,10 @@ use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
  */
 class SharingRole
 {
-    private string $id;
-    private string $displayName;
-    private string $description;
-    private int $weight;
+    private ?string $id;
+    private ?string $displayName;
+    private ?string $description;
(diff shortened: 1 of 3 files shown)


src/SharingRole.php
```diff
@@ -11,63 +11,64 @@
...
+    /**
+     * @param string|null|int $data
+     * @return string
+     */
+    public function validateData($data): string
+    {
+        return ($data === null || $data === '') ?
+        throw new InvalidResponseException(
+            "Invalid id returned for user '" . print_r($data, true) . "'"
+        ) : (string)$data;
```
individual-it, line 39: good try to reduce code, but now the message is not correct fro every case

individual-it (changes requested):

individual-it (approved):

individual-it:
There are minor code-style issues: https://drone.owncloud.com/owncloud/ocis-php-sdk/906/3/3
but ready to be merged when CI is green

PR #165, opened 2023-12-29
Adjust drone script

phil-davis:
1) name unit test pipelines correctly
2) name integration test steps correctly

I noticed these when trying to add PHP 8.3 to CI. We may as well fix this now.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -321,10 +321,10 @@ def tests(ctx, name, command, phpversions, coverage, xdebugNeeded = False):
     pipelines = []
     if name in config and config[name]:
         for php in phpversions:
-            name = "%s-%s" % (name, php)
+            fullname = "%s-%s" % (name, php)
             steps = cacheRestore() + [
                 {
-                    "name": name,
+                    "name": fullname,
                     "image": OC_CI_PHP % php,
                     "environment": {
                         "COMPOSER_HOME": "%s/.cache/composer" % dir["base"],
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -83,8 +83,8 @@ def main(ctx):
 
 def phpIntegrationTest(ctx, phpversions, coverage):
     pipelines = []
-    steps = keycloakService() + restoreOcisCache() + ocisService() + cacheRestore()
     for php in phpversions:
+        steps = keycloakService() + restoreOcisCache() + ocisService() + cacheRestore()
```
phil-davis, line 87: We need to start with `steps` set to just the initial steps at the start of each loop.
If we have multiple PHP versions to generate pipelines for, the previous code would end up with more-and-more steps being added on and get errors about duplicate step names.

.drone.star
```diff
@@ -321,10 +321,10 @@ def tests(ctx, name, command, phpversions, coverage, xdebugNeeded = False):
     pipelines = []
     if name in config and config[name]:
         for php in phpversions:
-            name = "%s-%s" % (name, php)
+            fullname = "%s-%s" % (name, php)
```
phil-davis, line 324: We don't want to change the value of `name` - if there are multiple PHP versions, then the previous code resulted in names ending in strings like `-8.1-8.2-8.3`

individual-it (approved):

PR #166, opened 2023-12-29
adjustment reflecting sharewithme changes in ocis

individual-it:
This PR makes adjustments to the SDK that are needed after https://github.com/owncloud/ocis/pull/8016
- make tests run with ocis wrapper (CI and locally)
- creates tests that check sharedWithMe when auto-accept of shares is disabled
- adjusts existing tests, so that they wait till the async auto-accept happened, and makes sure it does not check sharedWithMe too early
- for functions that do not return valid values if a share is not accepted (yet) throw `TooEarly` exception
- add function to check if a share is hidden
- add function to check if share is set to be syncronized


.drone.env
```diff
@@ -1,3 +1,3 @@
 # The version of OCIS to use in pipelines that test against OCIS
-OCIS_COMMITID=36b7db84b491e8a12c8520b8bf803c4175a09b1f
-OCIS_BRANCH=master
+OCIS_COMMITID=370eaacaf05d185ab305aa2c9841376fdd2b8230
+OCIS_BRANCH=graph-beta-fix-shared-with-me-to-test-with-php-sdk
```
individual-it, line 3: to be changed when https://github.com/owncloud/ocis/pull/8016 gets merged

src/ShareReceived.php
```diff
@@ -29,14 +30,12 @@ public function __construct(
...
+     * @throws TooEarlyException
      */
     public function getId(): string
     {
-        return empty($this->shareReceived->getId())
-            ? throw new InvalidResponseException(
-                "Invalid Id '" . print_r($this->shareReceived->getId(), true) . "'"
-            )
+        return ($this->shareReceived->getId() === null)
+            ? throw new TooEarlyException()
```
individual-it, line 38: Is that the correct exception? It might be that we are checking too-early and an auto-accept did not happen yet, or there is no auto-accept and the user has to manually accept the share AKA "set it to sync" in the webUI

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
```diff
@@ -174,4 +175,60 @@ protected function getContentOfResource425Save(OcisResource $resource): string
...
         return $content;
     }
+
+    /**
+     * wrapper around `getSharedWithMe()` that makes sure the share is auto-accepted
+     * the auto-accepting happens async, so calling getSharedWithMe() directly might be too early.
+     * @param Ocis $ocis
+     * @return array<ShareReceived>
+     */
+    protected function getSharedWithMeWaitTillShareIsAccepted(Ocis $ocis): array
```
individual-it, line 185: maybe we should call it something with "synced" as in the current webUI

tests/integration/Owncloud/OcisPhpSdk/ShareTestGetSharedWithMeTest.php
```diff
@@ -53,18 +56,28 @@ public function setUp(): void
...
-        //The step will only work after this bug is solved https://github.com/owncloud/ocis/issues/8000
-        //$this->assertSame($this->personalDrive->getId(), $receivedShare[0]->getDriveId());
+        $this->fileToShare->invite($this->einstein, $this->editorRole);
+        /**
+         * @var ShareReceived $receivedShare
+         */
+        $receivedShare = $this->getSharedWithMeWaitTillShareIsAccepted($this->einsteinOcis)[0];
+        $this->assertInstanceOf(ShareReceived::class, $receivedShare);
+        $this->assertMatchesRegularExpression('/' . $this->getUUIDv4Regex() . '/', $receivedShare->getRemoteItemId());
+        $this->assertMatchesRegularExpression('/' . $this->getFileIdRegex() . '/', $receivedShare->getId());
```
rhafer, line 68: Hm, all those IDs are supposed to be opaque. The fact that they are composed of UUIDs with some form of separators is an implementation detail. I am not exactly sure why a client implementation should test of that.

I haven't looked at the implementation of `getUUIDv4Regex()`, but there is something fishy about it. The `id` of the `remoteItem` is not a simple UUID. It's basically the same format as the driveItemId can be.

tests/integration/Owncloud/OcisPhpSdk/ShareTestGetSharedWithMeTest.php
```diff
@@ -53,18 +56,28 @@ public function setUp(): void
...
+        $this->assertSame($this->fileToShare->getName(), $receivedShare->getName());
+        $this->assertSame($this->fileToShare->getEtag(), $receivedShare->getEtag());
+        $this->assertSame($this->fileToShare->getId(), $receivedShare->getRemoteItemId());
+        $this->assertSame($this->fileToShare->getName(), $receivedShare->getRemoteItemName());
+        $this->assertSame($this->fileToShare->getSize(), $receivedShare->getRemoteItemSize());
+        $this->assertFalse($receivedShare->isUiHidden());
+        $this->assertEqualsWithDelta(time(), $receivedShare->getRemoteItemSharedDateTime()->getTimestamp(), 120);
+        $this->assertStringContainsString('Admin', $receivedShare->getOwnerName());
+        $this->assertMatchesRegularExpression('/' . $this->getUUIDv4Regex() . '/', $receivedShare->getOwnerId());
```
rhafer, line 77: The ID thing I wrote above is also true here.

rhafer (commented):
I can't really give a :+1:  or :-1: on this. Just add some thoughts. From what I understood of the code it looks mostly correct.

tests/integration/Owncloud/OcisPhpSdk/ShareTestGetSharedWithMeTest.php
individual-it: Thanks, I've changed the assertions to only check that the ID is not an empty string

amrita-shrestha (approved):
other than above queries, lgtm

tests/integration/Owncloud/OcisPhpSdk/OcisPhpSdkTestCase.php
individual-it: yes, this is only a function for tests

amrita-shrestha (approved):
lgtm

PR #169, opened 2024-01-04
Add more tests coverage for shares

amrita-shrestha:
Add tests for update Share Permission like password role and expiration

### DIFF
--- a/src/ShareCreated.php
+++ b/src/ShareCreated.php
@@ -3,6 +3,7 @@
 namespace Owncloud\OcisPhpSdk;
 
 use OpenAPI\Client\ApiException;
+use OpenAPI\Client\Model\Permission as ApiPermission;
 use OpenAPI\Client\Model\OdataError;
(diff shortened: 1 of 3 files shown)


src/ShareCreated.php
```diff
@@ -37,14 +38,14 @@ public function getPermissionId(): string
      */
     public function setRole(SharingRole $role): bool
     {
-        $this->apiPermission->setRoles([$role->getId()]);
-
+        $apiPermission = new ApiPermission();
+        $apiPermission->setRoles([$role->getId()]);
```
individual-it, line 42: why do we need this change?

amrita-shrestha: `$this->apiPermission` contains unnecessary values and `updatePermission` API endpoint body doesn't support those values so initiating new apiPermission which contain role key-value pair

individual-it (approved):

PR #170, opened 2024-01-05
refactoring notification class to validate id inside getters

nirajacharya2:
previously we validated the notification id from server in ocis class but now this PR moves that logic to the getters of Notification class.


individual-it (changes requested):
I know I have told you to do this change, but looking now at the code I don't like the approach.
In the other cases we get a whole object into the constructor and then check the values inside of it, here we get only the properties.
We really should not accept `mixed` for the id, because it cannot be `mixed`
I see two possibilities:
1. give `$ocsData` to the Nofication constructor and let the Notification class handle it all
2. leave it as it is

Issue #131, opened 2023-12-19
Throw proper error message while tests fails


amrita-shrestha:
Currently, error messages are not thrown if the integration test fails. Without an error message we have to debug more so i think proper messgae should be throen in some location in integration test

S-Panta:
Most of the function we use inside the source code all has their `proper error message` when the request fails. 
`Assert` statement has its style of just throwing assert Error.

S-Panta:
It would be good if the PR is made separately for each file inside the integration test to avoid more changes in a single PR.

- [ ] DriveTest
- [ ] GroupsTest
- [ ] NotificationTest
- [ ] OcisResoureceTest
- [ ] OcisTest
- [ ] ResoureceInviteTest
- [ ] ResourceShareLinkTest
- [ ] ShareCreatedModifyTest
- [ ] ShareGetSharedByMeTest
- [ ] ShareReceivedTest
- [ ] ShareTestGetSharedWithMeTest
- [ ] UsersTest

S-Panta:
Once all this PR get merged, I'll check over all the integration test for coherent message.

S-Panta:
Closing this PR as all is merged

PR #185, opened 2024-01-15
error message in assertion updated for ShareCreatedModifyTest

S-Panta:
Related issue : https://github.com/owncloud/ocis-php-sdk/issues/131


tests/integration/Owncloud/OcisPhpSdk/ShareCreatedModifyTest.php
```diff
@@ -54,16 +54,33 @@ public function testDeleteIndividualShare(): void
         $this->fileToShare->invite($this->marie, $this->viewerRole);
         $shares = $this->ocis->getSharedByMe();
         foreach ($shares as $share) {
-            $this->assertInstanceOf(ShareCreated::class, $share);
+            $this->assertInstanceOf(
+                ShareCreated::class,
+                $share,
+                "Expected class to be " . ShareCreated::class
```
individual-it, line 60: I think we can just write that class name in the string like:
```suggestion
                "Expected class to be 'ShareCreated'"
```

individual-it (changes requested):

individual-it (approved):

PR #187, opened 2024-01-16
Fix ocis cache check script

amrita-shrestha:
This PR changes the scripting file to check one commit ID once.
Checking both master and ocis commit id in the bash file makes it non-reusable for dynamic case.
In addition to that on CI,  `cache-ocis-master` and `cache-ocis-stable` run same bash script.


individual-it (commented):
1. need rebase and solving conflicts
2. I would suggest to have the variable renames into a separate PR

saw-jan (approved):
LGTM

PR #198, opened 2024-01-26
fix: rename isClientSynchronize method

phil-davis:
fix the spelling so that it is consistent with `getAtClientSynchronize` and other places where we spell Synchronize.

This will effect anything that uses this public method. I suppose that we will need to change any calls from the Moodle integration.

~And maybe it should be called `isClientSynchronized` to read more nicely?~ (done)

### DIFF
--- a/src/ShareReceived.php
+++ b/src/ShareReceived.php
@@ -32,10 +32,10 @@ class ShareReceived
      */
     public function getId(): ?string
     {
-        if (empty($this->shareReceived->getId()) && $this->isClientSyncronize() === false) {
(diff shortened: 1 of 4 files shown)


S-Panta (approved):

amrita-shrestha (approved):
lgtm

individual-it (changes requested):
@phil-davis if we are changing the name anyway then lets change it completely to `isClientSynchronized`

After this get's merged, the moodle plugin would need to get adjusted also https://github.com/owncloud/moodle-repository_ocis/blob/a4fd5829e33db2715d0726ef8d79a89adcf917fa/classes/ocis_manager.php#L621

Issue #200, opened 2024-01-29
Resolve ToDos in ocis.php related to OdataError


phil-davis:
https://github.com/owncloud/ocis-php-sdk/blob/main/src/Ocis.php has some ToDos about calls that might return an `OdataError` object.
```
        if ($allDrivesList instanceof OdataError) {
            // ToDo: understand how this can happen, and what to do about it.
            throw new InvalidResponseException(
                "listAllDrives returned an OdataError - " . $allDrivesList->getError()
            );
```

If something goes wrong in the underlying calls down the stack and across to the server, then libre-graph-api calls can return an `OdataError` object, rather than throwing `ApiException`.

Sort out what to do if this happens at run-time.

individual-it:
@phil-davis @amrita-shrestha is this done?

phil-davis:
yes, done in #201

PR #201, opened 2024-01-29
fix: improve handling of OdataError

phil-davis:
Issue #200 

If API calls return an OdataError object, then find the error code and message in the object, and use it to create and throw the appropriate HTTP error exception, like other code does. So consumers of the SDK can get a consistent type of exception when something goes wrong.

### DIFF
--- a/src/Exception/ExceptionHelper.php
+++ b/src/Exception/ExceptionHelper.php
@@ -4,6 +4,7 @@ namespace Owncloud\OcisPhpSdk\Exception;
 
 use GuzzleHttp\Exception\GuzzleException;
 use OpenAPI\Client\ApiException;
+use OpenAPI\Client\Model\OdataError;
 use Sabre\HTTP\ClientException as SabreClientException;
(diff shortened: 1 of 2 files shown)


src/Exception/ExceptionHelper.php
```diff
@@ -90,4 +91,40 @@ public static function getHttpErrorException(
...
+                $errorMessage,
+                (int) $errorCode
+            );
+            return self::getHttpErrorException($genericHttpException);
+        }
+        return new InvalidResponseException(
+            "$methodName returned an OdataError with code '" .
+            $errorCode .
+            "' and message '" . $errorMessage . "'"
+        );
```
amrita-shrestha, line 128: don't we have to return like above `error message, code and exception instance`
```
            500 => new InternalServerErrorException(
                $message,
                $e->getCode(),
                $e
            ),
```

phil-davis: Good idea. I will do that...

amrita-shrestha (approved):
lgtm

PR #207, opened 2024-02-16
clarify exception messages for webfinger issues

individual-it:
make it more explicit what failed when trying to extract data from the token


src/Ocis.php
```diff
@@ -234,22 +234,55 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
     {
         $tokenDataArray = explode(".", $this->accessToken);
         if (!array_key_exists(1, $tokenDataArray)) {
-            throw new \InvalidArgumentException(self::DECODE_TOKEN_ERROR_MESSAGE);
+            throw new \InvalidArgumentException(
+                self::DECODE_TOKEN_ERROR_MESSAGE .
+                " No payload found. Token: '" .
+                $this->accessToken .
```
phil-davis, line 240: Is it generally OK to return the token (in the exception message text) to the caller?
The caller provided the token in the first place, and it is invalid in these cases, so I suppose there is not any security problem.
The caller's code that handles exceptions might, for example, write the exception message to a log file. Handy for following-up problems, but maybe not so good if the caller is not aware of that the token  value is going in the log file.

phil-davis (approved):

PR #214, opened 2024-03-27
Cache latest commit ocis on nightly

amrita-shrestha:
# Description

1.This PR changes the way how ocis has been cached. From now on ocis will be cached in different way for nightly and PR builds. In nightly, the latest commit ID of ocis repo will be fetched and that ocis will be built. In PR builds, pinged ocis commit id will be used to cached.
2. Changed role related tests code to work with both master and stable branch of ocis . Assertion is some place to fail tests rather than throwing error
3. bumped latest ocis commit


amrita-shrestha:
block until this report marked as featured or bug https://github.com/owncloud/ocis/issues/8703#issuecomment-2036793521

amrita-shrestha:
ocis stable5.0 has share role different than master branch. So, we have to skip tests on master which uses shareRole. Or we can add workaround to map role for test

individual-it:
@amrita-shrestha what is exactly the difference? Does it affect the code or only the tests?

amrita-shrestha:
> @amrita-shrestha what is exactly the difference? Does it affect the code or only the tests?

@individual-it only test code

individual-it:
@amrita-shrestha then adjust our tests accordingly, so that they work with both versions on ocis

individual-it (approved):

PR #222, opened 2024-06-28
refactoring setmembers function in group.php and unit tests related to it

nirajacharya2:
function `setMembers` only had one line of code which was used in only one place so refactoring `setMembers` function in `Group.php` and unit tests related to it.

### DIFF
--- a/src/Group.php
+++ b/src/Group.php
@@ -124,15 +124,6 @@ class Group
     }
 
     /**
-     * Set the value of members
-     * @param User $member
-     */
-    public function setMembers(User $member): void
-    {
-        $this->members[] = $member;
-    }
-
-    /**
      * @param User $user
      *
      * @throws BadRequestException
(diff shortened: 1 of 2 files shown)


phil-davis:
Is there an issue about this?
It would be nice for future readers to know why this is being done.

src/Group.php
```diff
@@ -123,15 +123,6 @@ public function getMembers(): array
         return $this->members;
     }
 
-    /**
-     * Set the value of members
-     * @param User $member
-     */
-    public function setMembers(User $member): void
```
phil-davis, line 130: This was a public function, so it removes `setMembers` from the available external interface of this class.
Is anything using `setMembers` from the Moodle integration?
Is the functionality of `setMembers` a requirement for this SDK?

phil-davis (approved):

PR #224, opened 2024-07-02
Update libre graph package

amrita-shrestha:
This Pr update libre graph package which supports ocis 6.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "^8.1",
         "sabre/dav": "^4.6",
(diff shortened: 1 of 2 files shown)


phil-davis (approved):
LGTM - the tests pass, so this small code change must be compatible with ocis 5, because the tests are running with ocis 5 - am I correct?

amrita-shrestha:
> LGTM - the tests pass, so this small code change must be compatible with ocis 5, because the tests are running with ocis 5 - am I correct?

yes we run tests with ocis 5 as stable branch

PR #226, opened 2024-07-02
Add share invite method using root endpoint

amrita-shrestha:
This PR adds a function to create a share invite for the drive using the root endpoint.Environment for OCIS_VERSION is added in pipeline to skip this test on stable branch

Structure of file will be handled in later PR. DriveShare.php and Share.php will be handled when delete or setExpiration will add

### DIFF
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -598,19 +598,7 @@ class Drive
      */
     public function invite($recipient, SharingRole $role, ?\DateTimeImmutable $expiration = null): ShareCreated
     {
-        $driveItemInviteData = [];
-        $driveItemInviteData['recipients'] = [];
-        $recipientData = [];
-        $recipientData['object_id'] = $recipient->getId();
-        if ($recipient instanceof Group) {
-            $recipientData['at_libre_graph_recipient_type'] = "group";
-        }
-        $driveItemInviteData['recipients'][] = new DriveRecipient($recipientData);
-        $driveItemInviteData['roles'] = [$role->getId()];
-        if ($expiration !== null) {
-            $expirationMutable = \DateTime::createFromImmutable($expiration);
-            $driveItemInviteData['expiration_date_time'] = $expirationMutable;
-        }
+        $driveItemInviteData = OcisResource::getShareInviteBody($recipient, $role, $expiration);
 
         if (array_key_exists('drivesRootApi', $this->connectionConfig)) {
             $apiInstance = $this->connectionConfig['drivesRootApi'];
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -582,12 +582,11 @@ class Drive
 
     /**
      * Invite a user or group to the drive.
-     * Every recipient will result in an own ShareCreated object in the returned array.
      *
      * @param User|Group $recipient
      * @param SharingRole $role
      * @param \DateTimeImmutable|null $expiration
-     * @return ShareCreated
+     * @return DriveShare
      * @throws BadRequestException
      * @throws ForbiddenException
      * @throws HttpException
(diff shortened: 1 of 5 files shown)


tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -72,11 +72,51 @@ public function testDeleteEnabledDrive(): void
...
             $role,
             "Array contains not only 'SharingRole' items"
         );
     }
+
+    public function testCreateDriveInvite(): void
+    {
+        if(getenv('OCIS_VERSION') === "stable") {
+            $this->markTestSkipped();
+        };
```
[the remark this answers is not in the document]
S-Panta: Only stable and latest commits are used in the building the drone. It feels easy to draw env from the drone runtime of stable and master only.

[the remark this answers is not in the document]
individual-it: yes, but that is not the point. We skip the test on oCIS 5.0.0 but when in real-live s.o. runs the code with 5.0.0 it will crash unexpected. We need to catch that case

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -87,4 +87,46 @@ public function testGetDriveRole(): void
...
         );
     }
+
+    public function testCreateDriveInvite(): void
+    {
+        // At the time of writing, "stable" is major version 5 of ocis.
+        // This functionality works with major version 6.
+        // When ocis major version 6 has been released as "stable" then remove this test skip.
+        if (getenv('OCIS_VERSION') === "stable") {
+            $this->markTestSkipped(
```
phil-davis, line 97: You could also add a separate test that only runs on "stable" and is skipped if not "stable".

That test can do the "invite", and expect an exception (or whatever happens, I think that the non-existent endpoint might cause a 404 "not found" response, and the invite will generate an exception because of that)

Advantages will be:
- we test that the SDK code does respond appropriately when the endpoint is not there
- when "stable" changes from being v5 of ocis to v6 of ocis, that test will suddenly fail and we will automatically realize that we need to adjust the test suite

src/Drive.php
```diff
@@ -577,6 +580,70 @@ public function emptyTrashbin(): bool
         return true;
     }
 
+    /**
+     * Invite a user or group to the drive.
+     * Every recipient will result in an own ShareCreated object in the returned array.
```
phil-davis, line 585: This method returns a single `ShareCreated` object.
So I don't think that this comment is correct - the method does not return an array.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
phil-davis: We need test code (could be in a separate test case) to test what happens on ocis v5,
Something similar to what was done in PR #229 
`EndPointNotImplementedException` needs to be thrown if the ocis version does not have the endpoint available.

src/DriveShare.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
```
phil-davis, line 2:
```suggestion
namespace Owncloud\OcisPhpSdk;
```

Probably needs the namespace to be set.

src/Drive.php
```diff
@@ -577,6 +581,78 @@ public function emptyTrashbin(): bool
...
+        }
+        $driveItemInviteData['recipients'][] = new DriveRecipient($recipientData);
+        $driveItemInviteData['roles'] = [$role->getId()];
+        if ($expiration !== null) {
+            $expirationMutable = \DateTime::createFromImmutable($expiration);
+            $driveItemInviteData['expiration_date_time'] = $expirationMutable;
+        }
+
+        if (array_key_exists('drivesRootApi', $this->connectionConfig)) {
+            $apiInstance = $this->connectionConfig['drivesRootApi'];
```
S-Panta, line 617: These codes are also inside the resource invite function. I'm sure Sonar Cloud get pissed off with this repetition. https://sonarcloud.io/summary/new_code?id=owncloud_ocis-php-sdk&pullRequest=226

src/Drive.php
```diff
@@ -577,6 +580,81 @@ public function emptyTrashbin(): bool
...
+     * @throws HttpException
+     * @throws InvalidResponseException
+     * @throws NotFoundException
+     * @throws UnauthorizedException
+     * @throws InternalServerErrorException
+     */
+    public function invite($recipient, SharingRole $role, ?\DateTimeImmutable $expiration = null): DriveShare
+    {
+        if((version_compare($this->ocisVersion, '6.0.0', '<'))) {
+            throw new EndPointNotImplementedException(Ocis::ENDPOINT_NOT_IMPLEMENTED_ERROR_MESSAGE);
```
phil-davis, line 601: Check the "throws" in the PHPdoc above.
`EndPointNotImplementedException` needs to be added.

src/Drive.php
```diff
@@ -577,6 +580,81 @@ public function emptyTrashbin(): bool
...
+        };
```
phil-davis, line 602:
```suggestion
        }
```

src/DriveShare.php
```diff
@@ -0,0 +1,53 @@
...
+    public function getPermissionId(): string
+    {
+        $id = $this->apiPermission->getId();
+        if ($id === null || $id === '') {
+            throw new InvalidResponseException(
+                "Invalid id returned for permission '" . print_r($id, true) . "'"
+            );
+        }
+        return (string)$id;
```
phil-davis, line 35:
```suggestion
        return $id;
```

`getId` always returns `string` or `null` and we handled `null` above. So we already know that `$id` is a string.

phil-davis (approved):

Issue #223, opened 2024-07-02
Add support for drive root endpoint 


amrita-shrestha:
New endpoint has been added on ocis : https://owncloud.dev/libre-graph-api/#/drives.root/InviteSpaceRoot
Add code to support root endpoint api

- [x] update libre graph package https://github.com/owncloud/ocis-php-sdk/pull/224
- [x] get permission https://github.com/owncloud/ocis-php-sdk/pull/225
     - [x] add code which detected ocis version https://github.com/owncloud/ocis-php-sdk/pull/229 
     - [x] throw an exception about API endpoint not implemented on ocis version less than 6.0.0 using capabilities api 
- [x] share invite https://github.com/owncloud/ocis-php-sdk/pull/226
- [x] delete share https://github.com/owncloud/ocis-php-sdk/pull/227
- [x] update share https://github.com/owncloud/ocis-php-sdk/pull/230
- [x] https://github.com/owncloud/ocis-php-sdk/pull/232

## Ocis-php-sdk story-line for drive invitation
-  user A invite user B on drive `player`. 
- user A should be able to update the role, and expiration date and remove the user. 
- user B gets invited-drive on  getMYdrive.
- user B should able to get all resources inside invited drive.
- user B should able to invite other user if he/she has enough permission.
- user B should get permission of drive on which he/she invited 

## Test coverage
-  user send drive invitation to group
- user A invite user B on drive `player`. 
- user A should be able to update the role, and expiration date and remove the user. 
- user B gets invited-drive on  getMYdrive.
- user B should able to get all resources inside invited drive.
- user B should able to invite other user if he/she has enough permission.
- user B should get permission of drive on which he/she invited 

- [x] Check ocis-php-sdk with moodle-repository-ocis @nirajacharya2

phil-davis:
The endpoint is available in ocis major version 6, so the code needs to understand that.

amrita-shrestha:
@nirajacharya2 and @SagarGi split tests coverage

amrita-shrestha:
feature support for drive endpoint and tests has been added so closing this issue

PR #227, opened 2024-07-02
Add function to remove share invite using root endpoint

amrita-shrestha:
This PR refactor delete function of Share to work on both permission and root endpoint.

In this PR, ResourceShare has been introduced. Share class is parent class for DriveShare and ResourceShare. ResourceShare will be parent class for ShareLink, ShareCreated.
- added tests for deleting drive share on ocis version > 6.0.0. Test has been skipped for ocis version < 6.0.0

## Part of 
- https://github.com/owncloud/ocis-php-sdk/issues/223

### DIFF
--- a/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
@@ -170,4 +170,52 @@ class DriveTest extends OcisPhpSdkTestCase
             $this->fail("EndPointNotImplementedException was thrown unexpectedly");
         }
     }
+
+    /**
+     * @throws ForbiddenException
+     * @throws InvalidResponseException
+     * @throws BadRequestException
+     * @throws EndPointNotImplementedException
+     * @throws UnauthorizedException
+     * @throws HttpException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     * @throws \Exception
+     */
+    public function testDeleteDriveInvite(): void
+    {
+        $this->initUser('marie', 'radioactivity');
+
+        $marie = $this->ocis->getUsers('marie')[0];
+
+        $managerRole = null;
+
+        if (getenv('OCIS_VERSION') === "stable") {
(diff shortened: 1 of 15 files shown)


tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -170,4 +170,52 @@ public function testCreateDriveInvite(): void
...
+    public function testDeleteDriveInvite(): void
+    {
+        $this->initUser('marie', 'radioactivity');
+
+        $marie = $this->ocis->getUsers('marie')[0];
+
+        $managerRole = null;
+
+        if (getenv('OCIS_VERSION') === "stable") {
+            $this->markTestSkipped('Ocis version < 6.0.0 does not support creation of drive share invite so delete drive share test has been skipped');
```
phil-davis, line 194: This is correct. It is difficult, or "impossible", to test what happens in ocis V5 if someone tries to delete a drive share, because it is not possible to create a drive share in the first place.

Someone could try to delete a "bogus" drive share, putting some non-existent id. But that would be the same as any "completely wrong" attempt to use the non-existent end-point, and that is already covered by ocis tests.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -170,4 +170,52 @@ public function testCreateDriveInvite(): void
...
+                throw new \Error(
+                    "manager role not found "
+                );
+            }
+            // ocis stable doesn't support root endpoint
+            if (getenv('OCIS_VERSION') === "stable") {
+                $this->expectException(EndPointNotImplementedException::class);
+                $this->expectExceptionMessage("This method is not implemented in this ocis version");
+                $this->drive->invite($marie, $managerRole);
+            }
```
phil-davis, line 213: I don't think that this code can ever happen. There is already a check higher up that will have skipped the test.
This `if` statement can be deleted.

phil-davis (approved):

PR #225, opened 2024-07-02
Add support for get roles of drive using root endpoint

amrita-shrestha:
This PR adds function which gets permission list of drives using root endpoint 
- added tests coverage for getRole()

Part of issue #223

PR #229, opened 2024-07-04
added code for getting current oCis version

S-Panta:
New API endpoint could be added in a different version of ocis. oCIS will throw 404 status code , that leads to confusion, whether api request has problem or api endpoint has not implemented.
This PR adds capability api endpoint to get current version of oCIS so that it would be easy to dissect the methods to use in different version of oCIS


src/Ocis.php
```diff
@@ -306,6 +306,33 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
         }
         throw new InvalidResponseException('invalid webfinger response');
     }
+
+    /**
+     * returns current oCIS version
+     *
+     * @return string
```
phil-davis, line 313: A comment about what the version string looks like would be useful.
Is it usually like "5.0.5" or "v5.0.5" or?
That will be helpful for developers who make use of this method.

Maybe we should remove any "v" at the front, and always just return the 3-part semver version n.n.n - then it won't matter in future, if ocis gives the version with "v" or without "v".

phil-davis (commented):
Can we have an integration test for this?
Then we will know that it works, at least for the happy path.

src/Ocis.php
S-Panta: capability endpoint doesn't have v before the product version

phil-davis: ok, then they probably will not randomly add a "v" to the front.

src/Ocis.php
```diff
@@ -306,6 +306,28 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
...
+    /**
+     * returns current oCIS version in semantic versioning format ( eg "5.0.5" )
+     *
+     * @return string
+     * @throws InvalidResponseException
+     */
+    public function getOcisVersion(): string
+    {
+        $fullUrl = $this->getServiceUrl() . '/ocs/v1.php/cloud/capabilities';
+        $client = new Client($this->createGuzzleConfig($this->connectionConfig, $this->accessToken));
```
amrita-shrestha, line 319: maybe we could use https://github.com/owncloud/ocis-php-sdk/pull/229/files#diff-82277ffad27771b61ef747676ab6b941eff7ed78d53b2633a6e5c29fe202309bR102

src/Ocis.php
```diff
@@ -306,6 +306,28 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
         }
         throw new InvalidResponseException('invalid webfinger response');
     }
+
+    /**
+     * returns current oCIS version in semantic versioning format ( eg "5.0.5" )
```
amrita-shrestha, line 311:
```suggestion
     * returns current oCIS version in semantic versioning format ( e.g. "5.0.5" )
```

src/Ocis.php
```diff
@@ -306,6 +306,28 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
...
+     *
+     * @return string
+     * @throws InvalidResponseException
+     */
+    public function getOcisVersion(): string
```
amrita-shrestha, line 316: whats your plan to use this function?

individual-it:
Why does the function need to be public?
What about making it private and only using it internally when an OCIS object is initiated and setting a version in a private variable. Then when we contact a endpoint that we know is not available in that version, we can throw an exception

phil-davis:
> Why does the function need to be public?

IMO it does not need to be public, consumers of the SDK do not need to be able to get the ocis version, anything related to differences of the ocis version should be handled internally.

The function should be used to fetch and remember the ocis version.

@S-Panta you could also add code in this PR to do "throw an exception about API endpoint not implemented on ocis version less than 6.0.0 " for the existing code that was first introduced in PR #225 - that way the new `getOcisVersion` will be used.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -72,15 +72,13 @@ public function testDeleteEnabledDrive(): void
 
     public function testGetDriveRole(): void
     {
-        // At the time of writing, "stable" is major version 5 of ocis.
-        // This functionality works with major version 6.
-        // When ocis major version 6 has been released as "stable" then remove this test skip.
-        if (getenv('OCIS_VERSION') === "stable") {
+        try {
+            $role = $this->drive->getRoles();
+        } catch(\Exception) {
```
phil-davis, line 77:
```suggestion
        } catch(EndPointNotImplementedException) {
```

We can just catch the exception that we think could happen.
Then in the "catch" block,
`if (getenv('OCIS_VERSION') !== "stable")`
then fail the test.
Something like:
`$this->fail("EndPointNotImplementedException was thrown unexpectedly");`

And the `assertContainsOnlyInstancesOf` code will need to be put inside the `try` block - when we do get the `EndPointNotImplementedException` then we do not want to check the content of `$roles`

This would be a way to implement both test flows, for OCIS 5 and 6, within one test case.

Otherwise, write 2 completely separate test cases, and do the appropriate skipping so that only one test cazse runs in each pipeline.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -72,15 +73,16 @@ public function testDeleteEnabledDrive(): void
...
-        if (getenv('OCIS_VERSION') === "stable") {
+        try {
+            $role = $this->drive->getRoles();
+        } catch(EndPointNotImplementedException) {
+            if (getenv('OCIS_VERSION') !== "stable") {
+                $this->fail("EndPointNotImplementedException was thrown unexpectedly");
+            }
             $this->markTestSkipped(
                 'This test is skipped because root endpoint for drive share is not applicable for version 5 of OCIS.'
             );
```
phil-davis, line 84: We do not want to skip the test. At this point in the code we are running on "stable" and we got `EndPointNotImplementedException`. That is what the system is supposed to do. We can let the test finish "normally".

src/Ocis.php
```diff
@@ -306,6 +309,29 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
...
         throw new InvalidResponseException('invalid webfinger response');
     }
+
+    /**
+     * saves current oCIS version in semantic versioning format ( e.g. "5.0.5" )
+     *
+     * @return void
+     * @throws InvalidResponseException
+     */
+    private function getOcisVersion(): void
```
amrita-shrestha, line 319: if it set ocis version . how about calling setOcisVersion rather than getOcisVersion

src/Ocis.php
```diff
@@ -306,6 +309,29 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
...
+    {
+        $fullUrl = self::getServiceUrl() . '/ocs/v1.php/cloud/capabilities';
+        $response = $this->guzzle->request('GET', $fullUrl);
```
amrita-shrestha, line 322:
```suggestion
        $response = $this->guzzle->get($this->serviceUrl . '/ocs/v1.php/cloud/capabilities');
```

src/Ocis.php
phil-davis: naming is hard - `setOcisVersion` sounds like it could be called, and it will set ocis to some version (of course, it does not take any parameter, so a developer will see that you can't specify a version and force ocis to change to run that version)

Maybe it will be clearer if this method returns the current ocis version string. The code in the constructor will look more "normal". It will have something like:
`$this->ocisVersion = $this->getOcisVersion();`

phil-davis:
https://drone.owncloud.com/owncloud/ocis-php-sdk/1578/5/3
code-style problems. Please check code-style locally before pushing to GitHub.

src/Drive.php
```diff
@@ -583,9 +587,13 @@ public function emptyTrashbin(): bool
      * @throws UnauthorizedException
      * @throws InvalidResponseException
      * @throws InternalServerErrorException
+     * @throws EndPointNotImplementedException
      */
     public function getRoles(): array
     {
+        if((version_compare($this->ocisVersion, '6.0.0', '<='))) {
```
phil-davis, line 594: when was the endpoint implemented? 6.0.1?
maybe, in a comment, we should mention the exact version where it was implemented, to help future readers.

src/Ocis.php
```diff
@@ -306,6 +307,29 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
         }
         throw new InvalidResponseException('invalid webfinger response');
     }
+
+    /**
+     * saves current oCIS version in semantic versioning format ( e.g. "5.0.5" )
```
phil-davis, line 312:
```suggestion
     * returns the current oCIS version in semantic versioning format ( e.g. "5.0.5" )
```

src/Ocis.php
```diff
@@ -306,6 +307,29 @@ private function getServiceUrlFromWebfinger(string $webfingerUrl): string
...
+     *
+     * @return string
+     * @throws InvalidResponseException
+     */
+    private function getOcisVersion(): string
```
phil-davis, line 317: This function could internally save the ocis version the first time that it is called,
When it is called again, it can return the saved value, rather than always doing an API request.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
phil-davis: Or you can refactor the code to do the 2 different test flows more separately, or make 2 separate test cases.

Whatever is done, we want tests that pass in each environment.

amrita-shrestha:
private function getGuzzleMock(?string $responseContent = null): MockObject
    {
        if ($responseContent === null) {
            $responseContent = '
                {
                    "data": {
                        "capabilities": {
                            "core": {
                                "status": {
                                    "productversion": "6.0.0+aa6041abb6"
                               }
                            }
                        }
                    }
                }';
        }

        $streamMock = $this->createMock(StreamInterface::class);
        $streamMock->method('getContents')->willReturn($responseContent);
        $responseMock = $this->createMock(ResponseInterface::class);
        $responseMock->method('getBody')->willReturn($streamMock);
        $guzzleMock = $this->createMock(Client::class);
        $guzzleMock->method('get')
            ->willReturn($responseMock);
        return $guzzleMock;
    }

phil-davis (approved):

phil-davis:
Good - we can reuse a lot of this code when implementing the other root endpoints, where we need to know the ocis version etc.

PR #232, opened 2024-07-12
Refactor root support implementation

amrita-shrestha:
This PR include following changes 
- undo file structure changes done by https://github.com/owncloud/ocis-php-sdk/pull/227
- put all drive invitation related code on drive class
- update drive object after drive invite, update drive invitation, delete drive drive invitation


individual-it (commented):
generally looks fine, but is it possible to put the revert of #227 into a separate PR? That would make it easier to review

amrita-shrestha:
> generally looks fine, but is it possible to put the revert of #227 into a separate PR? That would make it easier to review



> generally looks fine, but is it possible to put the revert of #227 into a separate PR? That would make it easier to review

I don't see any point on doing that. 95% changes falls under recert PR and only 5% changes goes on new PR.

src/Drive.php
```diff
@@ -706,6 +722,144 @@ public function getRoles(): array
         return $roles;
     }
 
+    /**
+     * Permanently delete the current drive share
+     * $permissionId will be provided by getPermissionId()
+     * @param string $permissionId
+     * todo after delete drive object contain old permissions
```
individual-it, line 729: is this todo done?

src/Drive.php
```diff
@@ -706,6 +722,144 @@ public function getRoles(): array
...
+        $this->updateDriveObject();
+        return true;
+    }
+
+    /**
+     * get Permission id of specific user
+     *
+     * @throws NotFoundException
+     */
+    public function getPermissionId(string $userId): ?string
```
individual-it, line 758: 1. why limiting this to a specific user? I would like to query all users that have access to the drive. This way I would have to know first a userId and could ask if the user has access. But how to know what userId to ask about?
2. after getting the permissionId I cannot see a way to find out what that permission is allowed to do.
3. according to the example in https://owncloud.dev/libre-graph-api/#/drives.root/ListPermissionsSpaceRoot one user can have multiple permissions

What about having a function called `getPermissions` that returns an array of `DrivePermission`. The `DrivePermission` class can return things like userId and displayname and also the role etc. ?
The setters could be moved ether into that Class or stay inside the `Drive` class and take a complete object as parameter and not only the permission id

individual-it (changes requested):

phil-davis (approved):
LGTM - I don't see any problem. This should remove an backward-compatibility issues, and moves forward on inviting and managing for drive roots.

src/Drive.php
```diff
@@ -673,37 +685,174 @@ public function invite($recipient, SharingRole $role, ?\DateTimeImmutable $expir
...
+        if(version_compare($this->ocisVersion, self::MIN_OCIS_VERSION_DRIVE_INVITE, '<')) {
             throw new EndPointNotImplementedException(Ocis::ENDPOINT_NOT_IMPLEMENTED_ERROR_MESSAGE);
-        };
-        $guzzle = new Client(
-            Ocis::createGuzzleConfig($this->connectionConfig, $this->accessToken)
-        );
+        }
+        $apiRoles = $this->sendGetPermissionsRequest()->getAtLibreGraphPermissionsRolesAllowedValues();
+        if (empty($apiRoles)) {
+            throw new InvalidResponseException(
```
individual-it, line 693: good idea :+1: 
I thought about returning an empty array in that case, but really an exception is the correct way to do it as a drive should always have some permissions given to some user

individual-it (approved):
I've pushed a couple of commits to fix sonarcloud issues.
Please double check if this still works with the moodle extension and if yes, merge the PR

PR #234, opened 2024-07-18
tests for updating drive share role and inviting another user

S-Panta:
Tests added:
- testReceiverInviteOtherUserToDriveShare
- testReceiverInviteOtherUserToDriveShareWithNoInvitePermission
- testReceiverUpdatesDriveShareRole
- testReceiverUpdatesDriveShareRoleNoEnoughPermission
Part of https://github.com/owncloud/ocis-php-sdk/issues/223

### DIFF
--- a/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
+++ b/tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
@@ -591,41 +591,22 @@ class DriveTest extends OcisPhpSdkTestCase
         $marieOcis = $this->initUser('marie', 'radioactivity');
         $katherineOcis = $this->getOcis('katherine', 'gemini');
         $katherine = $this->ocis->getUsers('katherine')[0];
-
         $marie = $this->ocis->getUsers('marie')[0];
-
-        $managerRole = null;
-
+        // ocis stable doesn't support root endpoint
         if (getenv('OCIS_VERSION') === "stable") {
-            // ocis version < 6.0.0 doesn't support root endpoint so mocking getRoles values
-            $role = [
-                "id" => "312c0871-5ef7-4b3a-85b6-0e4074c64049",
-                "description" => "Allows managing a space",
-                "displayName" => "Manager",
-                "@libre.graph.weight" => 3
-            ];
-
-            $shareRoles = [new SharingRole(new UnifiedRoleDefinition($role))];
-        } else {
-            $shareRoles = $this->drive->getRoles();
+            $this->markTestSkipped("This method is not implemented in this ocis version");
         }
+        $shareRoles = $this->drive->getRoles();
         foreach ($shareRoles as $role) {
             if ($role->getId() === self::getPermissionsRoleIdByName('Manager')) {
                 $managerRole = $role;
                 break;
             }
         }
-
         if (empty($managerRole)) {
             throw new \Error(
                 "manager role not found "
             );
-        }
-        // ocis stable doesn't support root endpoint
-        if (getenv('OCIS_VERSION') === "stable") {
-            $this->expectException(EndPointNotImplementedException::class);
-            $this->expectExceptionMessage("This method is not implemented in this ocis version");
-            $this->drive->invite($marie, $managerRole);
         }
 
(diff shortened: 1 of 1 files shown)


tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -575,4 +575,307 @@ public function testGetMyDriveForDriveShareReceiver(): void
...
+            throw new \Error(
+                "manager role not found "
+            );
+        }
+        // ocis stable doesn't support root endpoint
+        if (getenv('OCIS_VERSION') === "stable") {
+            $this->expectException(EndPointNotImplementedException::class);
+            $this->expectExceptionMessage("This method is not implemented in this ocis version");
+            $this->drive->invite($marie, $managerRole);
+        }
```
amrita-shrestha, line 774: may be we can skip this part

individual-it: Yes, if it's tested somewhere else there is no point to copy it into every test

phil-davis: You will still need to do something to handle what happens when the unit test suite is run on old ocis. At least mark test skipped.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -575,4 +575,228 @@ public function testGetMyDriveForDriveShareReceiver(): void
...
+            $permissionId = $driveInvitation->getId();
+            if (empty($permissionId)) {
+                throw new \Error(" Permission not found of user Marie Curie");
+            }
+            $receivedInvitationDrive = $marieOcis->getDriveById($this->drive->getId());
+
+            $shareRoles = $receivedInvitationDrive->getRoles();
+
+            foreach ($shareRoles as $role) {
+                if ($role->getId() === self::getPermissionsRoleIdByName('Space Viewer')) {
```
phil-davis, line 732: What will happen if "Space Viewer" is somehow not one of the roles that is found?
I suppose that this loop will end with no assert happening. Probably the test will fail because PHPunit will see that the test did no assert.
But maybe it would be nicer to add some logic to detect that "nothing interesting happened" in this for loop, and fail in that case.
```
$spaceViewerRoleFound = false;
foreach { ...
  if () {
    code
    $spaceViewerRoleFound = true;
  }
}
$this->assertTrue($spaceViewerRoleFound, 'Space Viewer role was not found');
```

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -575,4 +575,228 @@ public function testGetMyDriveForDriveShareReceiver(): void
...
+                break;
+            }
+        }
+        if (empty($managerRole)) {
+            throw new \Error(
+                "manager role not found "
+            );
+        }
+
+
```
phil-davis, line 614:
```suggestion

```

Bonus empty line.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -575,4 +575,228 @@ public function testGetMyDriveForDriveShareReceiver(): void
...
+            }
+        }
+
+        if (empty($managerRole)) {
+            throw new \Error(
+                "manager role not found "
+            );
+        }
+
+
```
phil-davis, line 720:
```suggestion

```

Bonus empty line.

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -575,4 +575,228 @@ public function testGetMyDriveForDriveShareReceiver(): void
...
+            }
+        }
+        if (empty($viewerRole)) {
+            throw new \Error(
+                "manager role not found "
+            );
+        }
+
+
+
```
phil-davis, line 779:
```suggestion

```

Bonus empty lines.

phil-davis (approved):

PR #246, opened 2024-08-06
implementing listing education user

nirajacharya2:
This PR implements listing of an education user
https://owncloud.dev/libre-graph-api/#/educationUser/ListEducationUsers
https://owncloud.dev/libre-graph-api/#/educationUser/GetEducationUser

### DIFF
--- a/tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
+++ b/tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace unit\Owncloud\OcisPhpSdk;
+
+use Owncloud\OcisPhpSdk\EducationUser;
+use OpenAPI\Client\Model\EducationUser as EU;
+use Owncloud\OcisPhpSdk\Ocis;
+use PHPUnit\Framework\TestCase;
+
+class EducationUserTest extends TestCase
+{
+    /**
+     * @return void
+     */
+    public function testgetEducationUsers()
+    {
+        $ocis = $this->createMock(Ocis::class);
+        $ocis->method("getEducationUsers")->willReturn([
+            new EducationUser(new EU([
+                "id" => "id",
+                "display_name" => "displayname",
+                "mail" => "mail@mail.com",
+                "on_premises_sam_account_name" => "sd",
+            ]))
+        ]);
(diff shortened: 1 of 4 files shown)


phil-davis (commented):
Looks good - for the start we just support listing and getting the existing 4 user attributes, the same ones as we support for the ordinary User API.

We can have unit tests for the EducationUser class that will be very similar to the ones for the User class.

And add unit tests for the new getEducationUsers and getEducationUserById methods. Those will have to mock the actual AI responses etc.

Then we need to work out how to start up an ocis instance with working education user endpoints. For a start, we don't need to be able to write to those endpoints. We can just have a system that already has education users set up, and we can list and get them in integration tests.

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * @return void
+     */
+    public function testgetEducationUsers()
+    {
+        $ocis = $this->createMock(Ocis::class);
+        /** @phan-suppress-next-line PhanUndeclaredMethod */
+        $user = $ocis->getEducationUsers();
+        $this->assertContainsOnlyInstancesOf(EducationUser::class, $user);
```
phil-davis, line 19: what happens if `$user` is an empty array?
(if something went wrong and there are exactly 0 education users)

Maybe the test will pass.

For the unit tests, we will know exactly how many education users exist, so you can also do `assertCount` to check that the array of `EducationUser` has the expected size.

tests/unit/Owncloud/OcisPhpSdk/UserTest.php
```diff
@@ -4,7 +4,7 @@
 
 use OpenAPI\Client\Model\User;
 use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
-use Owncloud\OcisPhpSdk\User as SdkUser;
+use Owncloud\OcisPhpSdk\BaseUser as SdkUser;
```
phil-davis, line 7: Why do we need to change this?
`class User extends BaseUser`

So I think that the tests will pass without this change?

phil-davis:
`PhanDeprecatedImplicitNullableParam Implicit nullable parameters`
Those were fixed in #248 
Please rebase and get CI green again.

individual-it (approved):

PR #247, opened 2024-08-15
Coding style PER-CS2.0

phil-davis:
https://www.php-fig.org/per/coding-style/ extends PSR-12 and specifies the coding style for newer PHP features. See the link documentation for details.

We could require this style - it's to be discussed. This is what happens to code if we use PER-CS2.0 and let php-cs-fixer reformat the code.

The main differences seem to be:
- a space is required between a cast (like `(int)`) and the variable that it applies to.
- when there is a comma-separated list of things (function parameters, array items etc.) then the last item in the list must end with a comma. This syntax is valid in newish versions of PHP, it makes it easier to add another item to the list (minimizing the diff listing)
- empty classes can have the `{}` at the end of the class declaration `class MyException extends Exception {}`

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -5,5 +5,5 @@ $finder = PhpCsFixer\Finder::create()
 
 $config = new PhpCsFixer\Config();
 $config->setFinder($finder);
-$config->setRules(['@PSR12' => true]);
+$config->setRules(['@PER-CS2.0' => true]);
 return $config;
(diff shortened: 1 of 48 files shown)


phil-davis:
I'm putting this to draft for now. There is no urgency at all, and maybe we don't even want to do it.

individual-it:
The requirement was from the beginning PSR-12, but PER-CS2.0 `extends, expands and replaces PSR-12` so I would stick with PSR-12.

PR #249, opened 2024-08-21
implementing creating education user

nirajacharya2:
This PR implements creating education user and also adds a way to set bearer token for making API request related to education endpoint

### DIFF
--- a/src/BaseUser.php
+++ b/src/BaseUser.php
@@ -4,6 +4,7 @@ namespace Owncloud\OcisPhpSdk;
 
 use OpenAPI\Client\Model\User;
 use OpenAPI\Client\Model\EducationUser;
+use OpenAPI\Client\Model\ObjectIdentity;
 use Owncloud\OcisPhpSdk\Exception\InvalidResponseException;
 
 class BaseUser
@@ -12,6 +13,12 @@ class BaseUser
     private ?string $displayName;
     private ?string $mail;
     private ?string $onPremisesSamAccountName;
+    private ?string $surname;
+    private ?string $givenName;
+    /**
+     * @var array<ObjectIdentity>|null
+     */
+    private ?array $identities;
 
     /**
      * @param User|EducationUser $user
@@ -25,6 +32,9 @@ class BaseUser
         $this->displayName = $user->getDisplayName();
         $this->mail = $user->getMail();
         $this->onPremisesSamAccountName = $user->getOnPremisesSamAccountName();
+        $this->surname = $user->getSurname();
+        $this->givenName = $user->getGivenName();
+        $this->identities = $user->getIdentities();
     }
 
     /**
@@ -68,5 +78,30 @@ class BaseUser
     public function getOnPremisesSamAccountName(): string|null
     {
         return $this->onPremisesSamAccountName;
+    }
+
+    /**
+     * Get the value of surname
+     */
+    public function getSurname(): string|null
+    {
+        return $this->surname;
+    }
+
+    /**
+     * Get the value of givenName
+     */
+    public function getGivenName(): string|null
+    {
+        return $this->givenName;
+    }
+
+    /**
+     * Get the value of identities
+     * @return array<ObjectIdentity>|null
+     */
+    public function getIdentities(): ?array
+    {
+        return $this->identities;
     }
 }
--- a/src/EducationUser.php
+++ b/src/EducationUser.php
@@ -3,7 +3,23 @@
 namespace Owncloud\OcisPhpSdk;
 
 use Owncloud\OcisPhpSdk\BaseUser;
+use OpenAPI\Client\Model\EducationUser as EducationUserModel;
 
 class EducationUser extends BaseUser
 {
+    private ?string $primaryRole;
+
+    public function __construct(EducationUserModel $user)
+    {
+        parent::__construct($user);
+        $this->primaryRole = $user->getPrimaryRole();
+    }
+
+    /**
+     * Get the value of primaryRole
+     */
+    public function getPrimaryRole(): string|null
+    {
+        return $this->primaryRole;
+    }
 }
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -18,6 +18,8 @@ use OpenAPI\Client\Api\EducationUserApi;
 use OpenAPI\Client\ApiException;
 use OpenAPI\Client\Configuration;
 use OpenAPI\Client\Model\Drive as ApiDrive;
+use OpenAPI\Client\Model\EducationUser as EducationUserModel;
+use OpenAPI\Client\Model\ObjectIdentity;
 use OpenAPI\Client\Model\OdataError;
 use OpenAPI\Client\Model\Quota;
 use Owncloud\OcisPhpSdk\Exception\BadRequestException;
@@ -60,6 +62,7 @@ class Ocis
     private string $accessToken;
     private Configuration $graphApiConfig;
     private Client $guzzle;
+    private ?Client $educationGuzzle = null;
     private string $notificationsEndpoint = '/ocs/v2.php/apps/notifications/api/v1/notifications?format=json';
 
     /**
@@ -95,6 +98,7 @@ class Ocis
         string $serviceUrl,
         string $accessToken,
         array $connectionConfig = [],
+        ?string $educationAccessToken = null,
     ) {
         if (!self::isConnectionConfigValid($connectionConfig)) {
             throw new \InvalidArgumentException('Connection configuration is not valid');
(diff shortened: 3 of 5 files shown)


src/BaseUser.php
```diff
@@ -12,6 +13,12 @@ class BaseUser
     private ?string $displayName;
     private ?string $mail;
     private ?string $onPremisesSamAccountName;
+    private ?string $surName;
```
individual-it, line 16: I think its one word:

```suggestion
    private ?string $surname;
```

src/EducationUser.php
```diff
@@ -6,4 +6,19 @@
 
 class EducationUser extends BaseUser
 {
+    private ?string $primaryRole;
+
+    public function __construct(\OpenAPI\Client\Model\EducationUser $user)
```
individual-it, line 11: Everywhere else we have put that into the `use` block

src/Ocis.php
```diff
@@ -94,12 +98,14 @@ class Ocis
     public function __construct(
         string $serviceUrl,
         string $accessToken,
-        array $connectionConfig = []
+        array $connectionConfig = [],
+        string $educationAcessToken = ''
```
individual-it, line 102: what about making it to default to `null`?

src/Ocis.php
```diff
@@ -58,8 +60,10 @@ class Ocis
         'This method is not implemented in this ocis version';
     private string $serviceUrl;
     private string $accessToken;
+    private string $educationAcessToken;
```
individual-it, line 63: this variable is only used when creating `$educationGuzzle`, so I believe it can be removed

src/Ocis.php
```diff
@@ -58,8 +60,10 @@ class Ocis
...
     private Configuration $graphApiConfig;
     private Client $guzzle;
+    private Client $educationGuzzle;
```
individual-it, line 66: what about defaulting to `null`?

src/Ocis.php
```diff
@@ -722,6 +737,56 @@ public function getUserById(string $userId): User
...
+        ?string $surname = null,
+        ?string $givenName = null,
+        ?string $mail = null,
+        ?string $primaryRole = null,
+        ?string $issuer = null,
+        ?string $issuerAssignedId = null,
+    ): EducationUser {
+        if(isset($this->educationGuzzle)) {
+            throw new \InvalidArgumentException(
+                "Guzzle client for education user is not initilized."
```
individual-it, line 757: that is generally true, but the real reason why the client is not initialized is that there was no access token provided.
```suggestion
                "This function cannot be used because no authentication token was provided for the educationUser endpoints."
```

src/Ocis.php
individual-it: also `access` is written with two `c`

src/Ocis.php
```diff
@@ -115,13 +121,22 @@ public function __construct(
...
+            $this->educationGuzzle = new Client(self::createGuzzleConfig($connectionConfig, $this->educationAcessToken));
+        }
     }
 
     public function getServiceUrl(): string
     {
         return $this->serviceUrl;
     }
 
+    public function setEducationAcessToken(string $acessTocken): void
```
individual-it, line 134:
```suggestion
    public function setEducationAccessToken(string $acsessToken): void
```

src/Ocis.php
```diff
@@ -115,13 +121,22 @@ public function __construct(
         $this->graphApiConfig = Configuration::getDefaultConfiguration()->setHost(
             $this->serviceUrl . '/graph'
         );
+        if($educationAcessToken !== '') {
+            $this->educationGuzzle = new Client(self::createGuzzleConfig($connectionConfig, $this->educationAcessToken));
```
individual-it, line 125: just call the setter as you have it anyway

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -39,4 +39,29 @@ public function testListEducationUsers()
         $user = $ocis->getEducationUserById("user-id");
         $this->assertInstanceOf(EducationUser::class, $user);
     }
+
+    /**
+     * @return void
+     */
+    public function testCreateEducationUsers()
```
individual-it, line 46: I don't think this test does anything useful as you are mocking the exact method that you are testing.
you would have to mock `EducationUserApi::createEducationUser` and then call `$ocis->createEducationUser()` for real

individual-it (changes requested):
also more tests are needed e.g. to test the case when no token is provided

src/Ocis.php
phil-davis: `acsessToken` has a typo in the above suggestion.
I will check what happened...

tests/unit/Owncloud/OcisPhpSdk/UserTest.php
```diff
@@ -120,6 +128,7 @@ public function testInvalidUserData(array $data)
                 "display_name" => $data["display_name"],
                 "mail" => $data["mail"],
                 "on_premises_sam_account_name" => $data["on_premises_sam_account_name"],
+
```
phil-davis, line 131:
```suggestion
                "on_premises_sam_account_name" => $data["on_premises_sam_account_name"],
```

src/Ocis.php
```diff
@@ -115,13 +119,21 @@ public function __construct(
         $this->graphApiConfig = Configuration::getDefaultConfiguration()->setHost(
             $this->serviceUrl . '/graph'
         );
+        if($educationAccessToken !== null) {
```
phil-davis, line 122:
```suggestion
        if ($educationAccessToken !== null) {
```

phil-davis:
Please rebase and resolve conflicts that happened when #250 was merged.

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -2,41 +2,182 @@
...
+
+
+
+    /**
+     * @dataProvider educationUserData
+     *
+     * @param array<int, array<int, array<string, array<int, string>|string>>> $data
+     * @return void
+     */
+    public function testgetEducationUsers(array $data)
```
phil-davis, line 43:
```suggestion
    public function testGetEducationUsers(array $data)
```

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -2,41 +2,182 @@
...
+    /**
+     * @dataProvider educationUserData
+     *
+     * @param array<int, array<int, array<string, array<int, string>|string>>> $data
+     * @return void
+     */
+    public function testGetEducationUserWithOutToken(array $data)
+    {
+        $educationApi = $this->createMock(EducationUserApi::class);
+        $this->expectExceptionMessage("This function cannot be used because no authentication token was provided for the educationUser endpoints.");
```
individual-it, line 139:
```suggestion
        $this->expectExceptionMessage(
            "This function cannot be used because no authentication token was provided for the educationUser endpoints.
       ");
```

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -2,41 +2,182 @@
...
+    /**
+     * @dataProvider educationUserData
+     *
+     * @param array<int, array<int, array<string, array<int, string>|string>>> $data
+     * @return void
+     */
+    public function testListEducationUserWithOutToken(array $data)
+    {
+        $educationApi = $this->createMock(EducationUserApi::class);
+        $this->expectExceptionMessage("This function cannot be used because no authentication token was provided for the educationUser endpoints.");
```
individual-it, line 157: too long line, please shorten

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -2,41 +2,182 @@
...
+    /**
+     * @dataProvider educationUserData
+     *
+     * @param array<int, array<int, array<string, array<int, string>|string>>> $data
+     * @return void
+     */
+    public function testCreateEducationUserWithOutToken(array $data)
+    {
+        $educationApi = $this->createMock(EducationUserApi::class);
+        $this->expectExceptionMessage("This function cannot be used because no authentication token was provided for the educationUser endpoints.");
```
individual-it, line 175: too long line, please shorten

tests/unit/Owncloud/OcisPhpSdk/EducationUserTest.php
```diff
@@ -2,41 +2,182 @@
...
+        $this->expectException(InvalidResponseException::class);
+    }
+
+    /**
+     * @dataProvider educationUserData
+     *
+     * @param array<int, array<int, array<string, array<int, string>|string>>> $data
+     * @return void
+     */
+    public function testCreateEducationUserWithOutToken(array $data)
```
individual-it, line 172: without is one word

```suggestion
    public function testCreateEducationUserWithoutToken(array $data)
```

individual-it (changes requested):
looks generally good to me, except of small change requests.

But the next big thing is, that now we are in the situation that `Ocis` needs to have an `$accessToken`, but if used only for the education endpoints the developer might not have a personalized token to give to `$accessToken`.
So we have to make that not-mandatory and fix the functions that use it to not work if it does not exist.
But that can be done in a separate PR

individual-it (approved):

phil-davis (approved):

amrita-shrestha (approved):
lgtm

PR #251, opened 2024-08-23
chore: enforce php-cs-fixer trailing_comma_in_multiline

phil-davis:
In "modern" PHP, you can put a trailing comma at the end of "lists of things" such as an array, function parameter list, function call list, etc.

Without the trailing comma, you have something like:
```
[
  "a",
  "b",
  "c"
]
```

And when you want to add "d" to the list, you have to also put a comma after "c".
That makes extra lines in the diff listing, and so the "blame" history in Git is not as effective.
And it is a bit of cognitive load to think to add the comma after "c" then add "d". Although a modern IDE should do it for you these days.

The trailing comma is an optional thing, not specified either way in PSR-12. But we could decide to be consistent and always use the trailing comma.

These are the changes that php-cs-fixer makes.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -9,6 +9,7 @@ $config->setRules(
     [
         '@PSR12' => true,
         'single_space_around_construct' => true,
+        'trailing_comma_in_multiline' => ['elements' => ['arguments', 'arrays', 'parameters',]],
     ]
 );
 return $config;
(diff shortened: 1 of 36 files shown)


phil-davis:
This example of a diff made me think of this:
https://github.com/owncloud/ocis-php-sdk/pull/249/files#diff-82277ffad27771b61ef747676ab6b941eff7ed78d53b2633a6e5c29fe202309bR100-R101

phil-davis:
@individual-it @amrita-shrestha what do you think?
Is this a reasonable coding standard to use?

individual-it (approved):

PR #253, opened 2024-09-02
Add search api support

amrita-shrestha:
This PR adds two methods related to search API
- Search resources in a global scale (support various patterns to search)
- Search resources within drive

### DIFF
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -856,6 +856,42 @@ class Drive
     }
 
     /**
+     * Search resource within a drive
+     *
+     * @return array<OcisResource>
+     * @throws SabreClientException
+     * @throws UnauthorizedException
+     * @throws TooEarlyException
+     * @throws ForbiddenException
+     * @throws InvalidResponseException
+     * @throws HttpException
+     * @throws \DOMException
+     * @throws SabreClientHttpException
+     * @throws BadRequestException
+     * @throws ConflictException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     */
+    public function searchResource(string $pattern, ?string $limit = null): array
+    {
+        $pattern .= " scope:" . $this->getId();
+        $webDavClient = $this->createWebDavClient();
+        $webDavClient->setCustomSetting($this->connectionConfig, $this->accessToken);
+
+        $responses = $webDavClient->sendReportRequest($pattern, $limit, '/remote.php/dav/spaces');
+        $resources = [];
+        foreach ($responses as $response) {
+            $resources[] = new OcisResource(
+                $response,
+                $this->connectionConfig,
+                $this->serviceUrl,
+                $this->accessToken,
+            );
+        }
+        return $resources;
+    }
+
+    /**
      * @param array<string> $tags
      * @todo This function is not implemented yet! Place, name and signature of the function might change!
      */
--- a/src/Drive.php
+++ b/src/Drive.php
@@ -856,42 +856,6 @@ class Drive
     }
 
     /**
-     * Search resource within a drive
-     *
-     * @return array<OcisResource>
-     * @throws SabreClientException
-     * @throws UnauthorizedException
-     * @throws TooEarlyException
-     * @throws ForbiddenException
-     * @throws InvalidResponseException
-     * @throws HttpException
-     * @throws \DOMException
-     * @throws SabreClientHttpException
-     * @throws BadRequestException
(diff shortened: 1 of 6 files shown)


src/Drive.php
```diff
@@ -855,6 +855,42 @@ public function setPermissionExpiration(string $permissionId, ?\DateTimeImmutabl
...
+     * @throws InvalidResponseException
+     * @throws HttpException
+     * @throws \DOMException
+     * @throws SabreClientHttpException
+     * @throws BadRequestException
+     * @throws ConflictException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     */
+    public function searchResource(string $pattern, ?string $limit = null): array
```
individual-it, line 875: we should document what are valid values for `$pattern`

src/WebDavClient.php
```diff
@@ -150,4 +151,53 @@ public function setCustomSetting(array $connectionConfig, string $accessToken):
...
+     * @param string $pattern
+     * @param string|null $limit
+     * @param string|null $url
+     * @return array<string, array<int, array<string, string|null|object>>>
+     * @throws HttpException
+     * @throws SabreClientException
+     * @throws SabreClientHttpException
+     * @throws \DOMException
+     */
+    public function sendReportRequest(string $pattern, ?string $limit, ?string $url = ''): array
```
individual-it, line 165: Even if we need `$limit` as a string in the XML it feels weird to make this parameter in the function an string, I think it should be `int` and if the XML strictly needs a string, then cast it to string later.
Using `string` here will make it much more error-prone. Same applies to higher level functions

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -696,4 +696,38 @@ public function testReceiverUpdatesDriveShareRoleNotEnoughPermission(): void
...
+     * @throws HttpException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     * @throws \Exception
+     */
+    public function testSearchResourceInsideDrive(): void
+    {
+        $this->drive->createFolder('myfolder');
+        $this->drive->uploadFile('myfolder/somefile.txt', 'some content');
+        $this->drive->uploadFile('somefile.txt', 'root content');
```
individual-it, line 714: also upload files that should not be found, to make sure the function not just returns all files

tests/integration/Owncloud/OcisPhpSdk/DriveTest.php
```diff
@@ -696,4 +696,38 @@ public function testReceiverUpdatesDriveShareRoleNotEnoughPermission(): void
...
+            }
+            sleep(2);
+            $attempt++;
+        } while ($attempt < $maxAttempts);
+
+        $this->assertCount(
+            2,
+            $resources,
+            "Expected two resource but found " . count($resources),
+        );
```
individual-it, line 731: also check if the correct files have been returned

src/Drive.php
```diff
@@ -855,6 +855,42 @@ public function setPermissionExpiration(string $permissionId, ?\DateTimeImmutabl
...
+    {
+        $pattern .= " scope:" . $this->getId();
+        $webDavClient = $this->createWebDavClient();
+        $webDavClient->setCustomSetting($this->connectionConfig, $this->accessToken);
+
+        $responses = $webDavClient->sendReportRequest($pattern, $limit, '/remote.php/dav/spaces');
```
individual-it, line 881: why using `remote.php`? This is just historical route and should not be used in oCIS anymore

src/Ocis.php
```diff
@@ -1124,4 +1126,41 @@ public function getSharedByMe(): array
...
+     * @throws InvalidResponseException
+     * @throws HttpException
+     * @throws \DOMException
+     * @throws SabreClientHttpException
+     * @throws BadRequestException
+     * @throws ConflictException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     */
+    public function searchResource(string $pattern, ?string $limit = null): array
```
individual-it, line 1147: looks very similar to `Drive::searchResource`, can that be combined?

individual-it (changes requested):

src/Drive.php
amrita-shrestha: need **remote.php** for api remote

individual-it (approved):

PR #254, opened 2024-09-03
Add support for preview api

amrita-shrestha:
This PR adds a method to get a resource preview.
**getPreview** method is stored under **OcisResource** class
return binary data of the image


tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -501,4 +501,23 @@ public function testGetRolesOfDeletedResources(): void
...
+
+    public function testResourcePreview(): void
+    {
+        $path =  __DIR__ . "/../";
+        $imageData = \file_get_contents($path . '/filesForUpload/testavatar.jpg');
+        if ($imageData === false) {
+            throw new \InvalidArgumentException('Failed to read the file for upload.');
+        }
+        $this->personalDrive->uploadFile('testavatar.jpg', $imageData);
+        sleep(2);
```
individual-it, line 513: can we rather download the file and check for 425 status code?

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -501,4 +501,23 @@ public function testGetRolesOfDeletedResources(): void
...
+        $this->createdResources[$this->personalDrive->getId()][] = 'testavatar.jpg';
+        $resources = $this->personalDrive->getResources();
+        foreach ($resources as $resource) {
+            if ($resource->getName() === 'testavatar.jpg') {
+                $isPreviewValid = imagecreatefromstring($resource->getPreview('1024', '1024'));
```
individual-it, line 518: check also if the size of the returned image is correct
check also other sizes

src/OcisResource.php
```diff
@@ -558,4 +558,31 @@ private function getFileResponseInterface(string $fileId): ResponseInterface
...
+     * @throws BadRequestException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     * @throws InvalidResponseException
+     * @throws HttpException
+     */
+    public function getPreview(string $width, string $height): string
+    {
+        $webDavClient = new WebDavClient(['baseUri' => $this->serviceUrl . '/dav/spaces/']);
+        $webDavClient->setCustomSetting($this->connectionConfig, $this->accessToken);
```
individual-it, line 574: we are keep on repeating this two lines. Shouldn't we use `Drive::createWebDavClient` ?

individual-it (changes requested):

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
individual-it: 425 will not work, because its only for post-processing not for search indexing

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -501,4 +501,33 @@ public function testGetRolesOfDeletedResources(): void
...
+                $isPreviewValid = imagecreatefromstring($previewImageData);
+                $this->assertInstanceOf(\GdImage::class, $isPreviewValid, "The response contains invalid image data");
+
+                $imageInfo = getimagesizefromstring($previewImageData);
+                if ($imageInfo === false) {
+                    throw new \InvalidArgumentException('Preview image data has invalid data');
+                }
+                // ocis auto adjust size
+                $this->assertEquals(32, $imageInfo[0], "Expected width of preview image 32 but found {$imageInfo[0]}");
+                $this->assertEquals(16, $imageInfo[1], "Expected height of preview image 32 but found {$imageInfo[1]}");
```
individual-it, line 529:
```suggestion
                $this->assertEquals(16, $imageInfo[1], "Expected height of preview image 16 but found {$imageInfo[1]}");
```

src/OcisResource.php
individual-it: adding a function in OcisResource is also fine

individual-it (changes requested):

src/OcisResource.php
```diff
@@ -554,8 +561,35 @@ public function getContentStream()
...
     private function getFileResponseInterface(string $fileId): ResponseInterface
     {
-        $webDavClient = new WebDavClient(['baseUri' => $this->serviceUrl . '/dav/spaces/']);
-        $webDavClient->setCustomSetting($this->connectionConfig, $this->accessToken);
+        $webDavClient = $this->createWebDavClient();
         return $webDavClient->sendRequest("GET", $fileId);
     }
+
+    /**
+     * aspect-ratio of previews to be preserved even when ask for an unusual preview size in oCIS
```
individual-it, line 569:
```suggestion
     * the aspect-ratio of previews will be preserved even if not matching sizes are requested
```

src/OcisResource.php
```diff
@@ -554,8 +561,35 @@ public function getContentStream()
...
+     *
+     * @throws UnauthorizedException
+     * @throws ForbiddenException
+     * @throws BadRequestException
+     * @throws NotFoundException
+     * @throws InternalServerErrorException
+     * @throws InvalidResponseException
+     * @throws HttpException
+     */
+    public function getPreview(string $width, string $height): string
```
individual-it, line 579:
```suggestion
    public function getPreview(int $width, int $height): string
```
we don't want the user to be able to pass `ABC` here

individual-it (changes requested):
delete [m:\xdebug.log](https://github.com/owncloud/ocis-php-sdk/pull/254/files#diff-20942d5b9fd76c98ab90977117a0f7eda24503b263cfab5f68902fc424b4a71b) file

individual-it (approved):

PR #255, opened 2024-09-05
access token optional function added

S-Panta:
The access token is made optional because the education user can work without that token and has its own education access token. 
`setEducationGuzzleClient` is removed because the setter is not necessary for the SDK.  It can be set from outside by giving them in `educationAccessToken`
The new implementation is done in a way that if both token isn't provided, the error will be thrown in object creation.
 Each function that uses accesstoken is checked on their own function body.

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -59,7 +59,8 @@ class Ocis
     public const ENDPOINT_NOT_IMPLEMENTED_ERROR_MESSAGE =
         'This method is not implemented in this ocis version';
     private string $serviceUrl;
-    private string $accessToken;
+    private ?string $accessToken;
+    private ?string $educationAccessToken;
     private Configuration $graphApiConfig;
     private Client $guzzle;
     private ?Client $educationGuzzle = null;
@@ -96,7 +97,7 @@ class Ocis
      */
     public function __construct(
         string $serviceUrl,
-        string $accessToken,
+        ?string $accessToken = null,
         array $connectionConfig = [],
         ?string $educationAccessToken = null,
     ) {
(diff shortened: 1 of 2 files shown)


src/Ocis.php
```diff
@@ -217,7 +216,7 @@ public static function isConnectionConfigValid(array $connectionConfig): bool
      *         The function is public to make it testable.
      * @phpstan-param ConnectionConfig $connectionConfig
      */
-    public static function createGuzzleConfig(array $connectionConfig, string $accessToken): array
+    public static function createGuzzleConfig(array $connectionConfig, ?string $accessToken): array
```
individual-it, line 219: still, `accessToken` should never be null at this stage
```suggestion
    public static function createGuzzleConfig(array $connectionConfig, string $accessToken): array
```

src/Ocis.php
```diff
@@ -752,14 +776,14 @@ public function createEducationUser(
         ?string $issuerAssignedId = null,
         ?EducationUserApi $apiInstance = null,
     ): EducationUser {
-        if (!isset($this->educationGuzzle)) {
+        if ($this->educationAccessToken === null) {
```
individual-it, line 779: make a function similar to `checkIfAccessTokenExists`, or even reuse `checkIfAccessTokenExists`

individual-it (changes requested):
tests are missing for the new functionality!

S-Panta:
Drone CI error in php stan 8.1 is extremely weird. https://drone.owncloud.com/owncloud/ocis-php-sdk/1800/7/3
I run php-phan in PHP 8.1 both in my local setup and local drone ci setup and there is no error.
The line number also doesn't match in this CI error

individual-it (approved):

PR #263, opened 2024-09-16
refractoring `createEducationUser` function

nirajacharya2:
This PR updated `createEducationUser` function and makes few optional parameters required parameters

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -781,12 +781,12 @@ class Ocis
     public function createEducationUser(
         string $displayName,
         string $onPremisesSAMAccountName,
+        string $issuer,
+        string $issuerAssignedId,
+        string $primaryRole,
         ?string $surname = null,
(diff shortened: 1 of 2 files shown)


amrita-shrestha (approved):
lgtm

src/Ocis.php
```diff
@@ -804,7 +804,7 @@ public function createEducationUser(
             "mail" => $mail,
             "on_premises_sam_account_name" => $onPremisesSAMAccountName,
             "primary_role" => $primaryRole,
-            "identities" => $identities,
+            "identities" => [$identities],
```
phil-davis, line 807: It looks like there can only be one identity at this point, which is the object create by `new ObjectIdentity`
So the variable name could be singular:
`"identities" => [$identity],`

phil-davis: See PR #264

Issue #260, opened 2024-09-11
Support for education endpoint


amrita-shrestha:
- [ ] creating education user https://github.com/owncloud/ocis-php-sdk/pull/249
- [ ] delete education user https://github.com/owncloud/ocis-php-sdk/pull/259
- [ ] https://github.com/owncloud/ocis-php-sdk/pull/255
- [ ] list education user https://github.com/owncloud/ocis-php-sdk/pull/246

PR #265, opened 2024-09-23
adding documentation for using education endpoint

nirajacharya2:
This PR adds documentation for using education endpoint

related issue: https://github.com/owncloud/ocis-php-sdk/issues/260

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,16 +40,29 @@ composer require owncloud/ocis-php-sdk
 ```
 
 ## Getting started
-Create an Ocis object using the service Url and an access token:
+Ocis has two types of access token, one which is used to interact with drive, group, user, shares etc.(OICD access token) and another which can be used to interact with education endpoints (education access token).
+
+Create an Ocis object using the service Url and an OIDC access token:
 ```php
 $ocis = new Ocis('https://example.ocis.com', $accessToken);
 ```
+Or create an Ocis object to interact with education endpoints using the service Url and an education access token:
+```php
+$ocis = new Ocis('https://education.ocis.com', null, [], $educationAccessToken);
+```
 
-Acquiring an access token is out of scope of this SDK, but you can find [examples for that below](#acquiring-an-access-token).
+At least one access token should be provided to use the SKD.
 
-Also refreshing tokens is not part of the SDK, but after you got a new token, you can update the Ocis object:
+Acquiring an OICD access token is out of scope of this SDK, but you can find [examples for that below](#acquiring-an-access-token).
+
+Also refreshing OICD tokens is not part of the SDK, but after you got a new token, you can update the Ocis object:
 ```php
 $ocis->setAccessToken($newAccessToken);
+```
+
+Education access token is set when starting the ocis graph server. You can get it using following snippet
+```php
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,8 @@ composer require owncloud/ocis-php-sdk
 Create an Ocis object using the service Url and an access token:
 ```php
 $ocis = new Ocis('https://example.ocis.com', $accessToken);
+// for using eduaction endpoint use education api token
```
phil-davis, line 46:
```suggestion
// for using education endpoint use education api token
```

README.md
```diff
@@ -43,6 +43,8 @@ composer require owncloud/ocis-php-sdk
 Create an Ocis object using the service Url and an access token:
 ```php
 $ocis = new Ocis('https://example.ocis.com', $accessToken);
+// for using education endpoint use education api token
+$ocis = new Ocis('https://education.ocis.com', null, [], $educationAccessToken);
 ```
```
individual-it, line 48: I think this is a bit confusing.
On top we are talking about an `access token`
In between there is an `education api token` but the variable is called `$educationAccessToken`
And then afterwards we tell people how to get an access token.

I think you will have to elaborate about
1. that there are two different tokens for different API endpoints
2. that at least one needs to be provided
3. where to get an education access token
4. which token is needed for which functions
5. which token the chapter below is talking about

individual-it (changes requested):

README.md
```diff
@@ -40,18 +40,31 @@ composer require owncloud/ocis-php-sdk
...
 
 ## Getting started
-Create an Ocis object using the service Url and an access token:
+Ocis has two type of access token, One which is used to interact with drive, group, user, shares etc.(OICD access token) and another which can be used to interact with education endpoints (education access token).
```
individual-it, line 43:
```suggestion
Ocis has two types of access tokens, one which is used to interact with drive, group, user, shares etc.(OICD access token) and another one which can be used to interact with education endpoints (education access token).
```

README.md
```diff
@@ -40,18 +40,31 @@ composer require owncloud/ocis-php-sdk
...
+
+Create an Ocis object using the service Url and an OIDC access token:
 ```php
 $ocis = new Ocis('https://example.ocis.com', $accessToken);
 ```
+Or create an Ocis object to interact with education endpoint using the service Url and an education access token:
```
individual-it, line 49:
```suggestion
Or create an Ocis object to interact with education endpoints using the service Url and an education access token:
```

README.md
```diff
@@ -40,18 +40,31 @@ composer require owncloud/ocis-php-sdk
...
+```php
+$ocis = new Ocis('https://education.ocis.com', null, [], $educationAccessToken);
+```
 
-Acquiring an access token is out of scope of this SDK, but you can find [examples for that below](#acquiring-an-access-token).
+Atleast one access token should be provided to use the sdk.
```
individual-it, line 54:
```suggestion
At least one access token should be provided to use the SDK.
```

individual-it (changes requested):

individual-it (approved):

PR #273, opened 2024-12-10
implement get resource by path

amrita-shrestha:
This PR implements a function to get resource 

Needed to use in https://github.com/JankariTech/DropboxExtensionForJoomla/pull/270

Updated ocis commit id


src/Drive.php
```diff
@@ -463,11 +445,56 @@ public function createFolder(string $path): bool
...
 
     /**
-     * @todo This function is not implemented yet! Place, name and signature of the function might change!
+     * send propfind request
+     * @param string $path
+     *
+     * @return array<OcisResource>
      */
-    public function getResourceMetadata(string $path = "/"): \stdClass
+    public function makePropfindRequest(string $path = "/"): array
```
individual-it, line 453:
```suggestion
    private function makePropfindRequest(string $path = "/"): array
```

individual-it (changes requested):
please also create tests for:
- empty folder
- folder with content
- non existing file/folder

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -530,4 +530,61 @@ public function testResourcePreview(): void
             }
         }
     }
+
+    public function testFileGetResourcesMetadata(): void
+    {
+        $resourceMetadata = $this->personalDrive->getResourceMetadata('/somefile.txt');
+        $driveResources = $this->personalDrive->getResources();
+        foreach ($driveResources as $driveResource) {
+            if ($driveResource->getName() === $resourceMetadata['name']) {
```
individual-it, line 539: I think its better to make it explicit, to make sure there is not a parallel bug in both `getName` and `$resourceMetadata['name']`
```suggestion
            if ($driveResource->getName() === 'somefile.txt') {
```

same thing for the other tests

tests/integration/Owncloud/OcisPhpSdk/OcisResourceTest.php
```diff
@@ -530,4 +530,61 @@ public function testResourcePreview(): void
...
+    /**
+     * @dataProvider invalidResources
+     */
+    public function testNonExistingResourceGetResourcesMetadata(string $invalidResources): void
+    {
+        $this->expectException(NotFoundException::class);
+        $this->personalDrive->getResourceMetadata($invalidResources);
+    }
+
+    public function testFolderGetResourcesMetadata(): void
```
individual-it, line 573: I think this test is :snake: :oil_drum: 
In a PROPFIND to a folder, the element 0 is the folder itself, that is thrown away in https://github.com/owncloud/ocis-php-sdk/pull/273/files#diff-326a1d6af0eef9fa2f105cb30f9476da1321c102f02aa3482b2d1983b8ff12e7R388
The new element 0 is the first child item. I believe `getResourceMetadata('/subfolder')` will get the properties of `uploaded.txt` or `innerfolder`
The test still passes because while cycling through the resources it comes to `uploaded.txt` and that is the same as `$resourceMetadata['name']`
I expect this test to fail when you do an explicit comparison as suggested by me above

individual-it (changes requested):

individual-it (approved):

PR #291, opened 2026-02-16
chore: build ocis master with Go 1.25

nirajacharya2:
using go 1.25 when building ocis master because of -> https://drone.owncloud.com/owncloud/ocis-php-sdk/2537/2/5


.drone.star
```diff
@@ -207,10 +207,14 @@ def buildOcis(branch):
     ocis_commit_id = getCommitId(branch)
     ocis_branch = getBranchName(branch)
     ocis_repo_url = "https://github.com/owncloud/ocis.git"
+    if branch == "master":
+        goImage = "owncloudci/golang:1.25"
```
individual-it: let's put that also into a const up the top something like `OC_CI_GOLANG_MASTER`

individual-it (changes requested):

individual-it (approved):

PR #294, opened 2026-03-26
feat(ci): migrate `php checks` , `docs`, and `php unit tests` from Drone to GitHub Actions

Ashim-Stha:
## Description
Migrates the following Drone CI pipelines to GitHub Actions:

- `php-checks` - codestyle, phpstan, phan
- `unit-tests-without-coverage` 
- `unit-tests-with-coverage` 
- `docs`

## Related Issue
- Part of: https://github.com/owncloud/ocis-php-sdk/issues/295

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,111 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+env:
+  COMPOSER_HOME: ${{ github.workspace }}/.cache/composer
+  DEFAULT_PHP_VERSION: '8.1'
+
+jobs:
+  php-checks:
+    name: PHP checks
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f # v2.37.0
+        with:
+          php-version: ${{ env.DEFAULT_PHP_VERSION }}
+          tools: composer:72a8f8e653710e18d83e5dd531eb5a71fc3223e6 # v2.9.5
+
+      - uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: .cache/composer
+          key: composer-${{ hashFiles('composer.lock') }}
+
+      - run: composer install
+
+      - name: codeStyle
+        run: make test-php-style
+
+      - name: phpStan
+        run: make test-php-phpstan
+
+      - name: phpPhan
+        run: make test-php-phan
+
+  php-unit-tests:
+    name: php unit tests ${{ matrix.php-version }} with coverage
+    runs-on: ubuntu-latest
+    needs: php-checks
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version : ['8.1', '8.2', '8.3']
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f # v2.37.0
+        with:
+          php-version: ${{ matrix.php-version }}
+          tools: composer:72a8f8e653710e18d83e5dd531eb5a71fc3223e6 # v2.9.5
+          coverage: xdebug
+
+      - uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: .cache/composer
+          key: composer-${{ hashFiles('composer.lock') }}
+
+      - run: composer install
+
+      - name: unitTests-${{ matrix.php-version }}
+        run: make test-php-unit
+
+      - name: coverage rename
+        run: mv tests/output/clover.xml tests/output/clover-unitTests-${{ matrix.php-version }}.xml
+
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: coverage-unitTests-${{ matrix.php-version }}
+          path: tests/output/clover-unitTests-${{ matrix.php-version }}.xml
+
+  docs:
+    name: docs
+    runs-on: ubuntu-latest
+    needs: php-unit-tests
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: docs generate
+        run: docker run --rm -v ${{ github.workspace }}:/data phpdoc/phpdoc@sha256:2b36e4f74937e40246d54ed12dfa4f988f21ed8d19e46f426ac3dc32eac2ff1d # v3.9.1
+
+      - name: publish api docs
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        if: github.event_name != 'pull_request'
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs
+          publish_branch: docs
+          force_orphan: true
+
+      - name: compile docs hugo
+        run: |
+          mkdir docs-hugo
+          cat docs-hugo-header.md README.md > docs-hugo/_index.md
+
+      - name: publish docs hugo
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        if: github.event_name != 'pull_request'
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: docs-hugo
+          publish_branch: docs-hugo
+          force_orphan: true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+  COMPOSER_HOME: ${{ github.workspace }}/.cache/composer
+  DEFAULT_PHP_VERSION: '8.1'
+
+jobs:
+  php-checks:
+    name: PHP checks 8.1
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
phil-davis:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

ownCloud wants to use the SHA of the commit, rather than a version tag. The version tag can be changed by a bad actor.

See examples in PR https://github.com/owncloud/reusable-workflows/pull/22

And `actions/checkout` is at major release v6.

All the `uses:` steps need to be modified to use the SHA of the commit id that is their current release.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+          path: tests/output/clover-unitTests-${{ matrix.php-version }}.xml
+
+  docs:
+    name: docs
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f # v2.37.0
```
saw-jan: most commonly used php setup but not verified.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+        with:
+          path: .cache/composer
+          key: composer-${{ hashFiles('composer.lock') }}
+
+      - run: composer install
+
+      - name: docs generate
+        uses: docker://phpdoc/phpdoc:3
```
saw-jan: using docker image. needs to clarify if this is ok.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: publish api docs
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
```
saw-jan, line 91: from unverified creator

saw-jan (commented):
Referencing:
>:white_check_mark: Allowed Actions Policy
>
>Only the following GitHub Actions are allowed:
>
>- Official GitHub Actions
>- Actions from verified publishers
>
>Any exceptions must be discussed with the development team before use.

we need to confirm (with dev team) that we are okay to use these actions

Ashim-Stha:
what will happen to owncloud-ci images?  we are using ubuntu and php images from marketplace
cc: @phil-davis , @DeepDiver1975

.github/workflows/ci.yml
Ashim-Stha: cc: @DeepDiver1975

.github/workflows/ci.yml
Ashim-Stha: cc: @DeepDiver1975

.github/workflows/ci.yml
Ashim-Stha: cc: @DeepDiver1975

phil-davis: See https://github.com/owncloud/reusable-workflows/blob/main/.github/workflows/php-unit.yml
and any other workflows in that folder. Those are being used with ownCloud core and apps.
Use anything from there as examples.

I don't think that we will be using https://github.com/owncloud-ci/php because actually we are easily using shivammathur/setup-php to get and configure PHP.

phil-davis: github.com/owncloud/impersonate/pull/256 has GitHub workflow actions that I am doing for ownCloud core and apps. So look there also to get any ideas.

.github/workflows/ci.yml
DeepDiver1975: please see if any of these fits the needs https://github.com/marketplace?query=phpdoc&type=actions

.github/workflows/ci.yml
DeepDiver1975: https://github.com/marketplace/actions/github-pages-action

DeepDiver1975:
> what will happen to owncloud-ci images

generally speaking: we will no longer need them. but there are exceptions like the squish ci image. we decide as we walk

.github/workflows/ci.yml
```diff
@@ -0,0 +1,149 @@
...
+        with:
+          path: .cache/composer
+          key: composer-${{ hashFiles('composer.lock') }}
+
+      - run: composer install
+
+      - name: unitTests-${{ matrix.php-version }}
+        run: make test-php-unit
+
+  php-unit-tests-with-coverage:
```
DeepDiver1975: you can simplify this and add the coverage configuration to the matrix

.github/workflows/ci.yml
Ashim-Stha: > https://github.com/marketplace/actions/github-pages-action

I’m a bit confused by the comment since the Marketplace link points to the same `peaceiris/actions-gh-pages` action I am already using. Should I use the same action? @DeepDiver1975

.github/workflows/ci.yml
Ashim-Stha: > please see if any of these fits the needs https://github.com/marketplace?query=phpdoc&type=actions

I checked the Marketplace options, but the first one I found appears to be very similar to what we have now (it still uses a Docker-based phpDoc image, e.g. docker://phpdoc/phar-ga:master). Should I use the same? @DeepDiver1975

.github/workflows/ci.yml
Ashim-Stha: we are not allowed to use `shivammathur/setup-php` now @phil-davis @DeepDiver1975 
```
[Error](https://github.com/owncloud/ocis-php-sdk/actions/runs/23892497636/workflow)
The actions shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f, docker://phpdoc/phpdoc:3, and peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e are not allowed in owncloud/ocis-php-sdk because all actions must be from a repository owned by owncloud, created by GitHub, or verified in the GitHub Marketplace. All actions must also be pinned to a full-length commit SHA.
```

phil-davis: I think that should be allowed now, it runs for me in `owncloud/impersonate`

DeepDiver1975 (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        coverage: ['with-coverage', 'without-coverage']
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f # v2.37.0
+        with:
+          php-version: ${{ matrix.php-version }}
+          tools: composer:72a8f8e653710e18d83e5dd531eb5a71fc3223e6 # v2.9.5
+          coverage: xdebug
```
DeepDiver1975, line 59: xdebug is active now on all configurations - coverage will be generated in all cases.

You want to remove this for non-coverage runs.

DeepDiver1975: But looking at the execution times this of 0 matter - just generate the coverage in all cases

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: make test-php-style
+
+      - name: phpStan
+        run: make test-php-phpstan
+
+      - name: phpPhan
+        run: make test-php-phan
+
+  php-unit-tests:
+    name: php unit tests ${{ matrix.php-version}} ${{ matrix.coverage }}
```
DeepDiver1975: please add a dependency to the php-checks step - there is no need to run unit tests if checks are failing

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: shivammathur/setup-php@accd6127cb78bee3e8082180cb391013d204ef9f # v2.37.0
+        with:
+          php-version: ${{ env.DEFAULT_PHP_VERSION }}
+          tools: composer:72a8f8e653710e18d83e5dd531eb5a71fc3223e6 # v2.9.5
+          coverage: none
+
+      - uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
```
DeepDiver1975: needed?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        with:
+          path: .cache/composer
+          key: composer-${{ hashFiles('composer.lock') }}
+
+      - run: composer install
```
DeepDiver1975: needed?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: mv tests/output/clover.xml tests/output/clover-unitTests-${{ matrix.php-version }}.xml
+
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        if: matrix.coverage == 'with-coverage'
+        with:
+          name: coverage-unitTests-${{ matrix.php-version }}
+          path: tests/output/clover-unitTests-${{ matrix.php-version }}.xml
+
+  docs:
+    name: docs
```
DeepDiver1975, line 80: we should only release docs if all tests are passing - right?

Ashim-Stha: Yes, I will add integeration tests dependecy too in another PR https://github.com/owncloud/ocis-php-sdk/pull/296 after implementing integration tests

DeepDiver1975 (approved):

PR #311, opened 2026-06-29
ci: adjust code and tests for phpstan v2

phil-davis:
Most of the changes to tests are just to ignore `phpstan` messages saying that the assertion is not necessary. I left those assertions in place, because, in the tests, they are double-checking the declared/expected types.

Most of the real source codes changes are minor adjustments or PHPdoc changes so that `phpstan` is happy. See a couple of comments below for things that look more interesting.

IMO Sonar Analysis is “not a thing” any more. Ignore that failing.

This will replace the Dependabot PR #303

### DIFF
--- a/src/Ocis.php
+++ b/src/Ocis.php
@@ -242,6 +242,7 @@ class Ocis
     /**
      * check for access token.
      *
+     * @phpstan-assert string $this->accessToken
      * @throws \InvalidArgumentException
      */
     private function checkIfAccessTokenExists(): void
@@ -256,6 +257,7 @@ class Ocis
     /**
      * check for access token of education user.
      *
+     * @phpstan-assert string $this->educationAccessToken
      * @throws \InvalidArgumentException
      */
     private function checkIfEducationAccessTokenExists(): void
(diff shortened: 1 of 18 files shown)


src/Ocis.php
```diff
@@ -960,7 +962,7 @@ public function getEducationSchools(?EducationSchoolApi $apiInstance = null): ar
      */
     public function getEducationSchoolById(string $schoolId, ?EducationSchoolApi $apiInstance = null): EducationSchool
     {
-        $this->checkIfAccessTokenExists();
+        $this->checkIfEducationAccessTokenExists();
```
phil-davis, line 965: Note: this was actually the wrong checking call, which I have fixed here.

src/WebDavClient.php
```diff
@@ -100,10 +101,7 @@ private function getCurlProxySettings(array $connectionConfig): array
 
             if (is_string($connectionConfig['proxy'])) {
                 $settings[CURLOPT_PROXY] = $connectionConfig['proxy'];
-            } elseif (
-                array_key_exists('proxy', $connectionConfig) &&
```
phil-davis, line 104: We already know that this array key exists from the `if` at line 99.
`phpstan` told me.

nirajacharya2 (approved):

